PR #3, opened 2025-10-07
add basic stuff for Github

mvadari:


mvadari:
> Shouldn't we keep the README file for now?

This was a duplicate README - there was one at the top level and one in `/xrpl-wasm-std`.

PR #34, opened 2025-10-21
Fix: u64 field size mismatch in cached ledger object getter

tekvyy:

Fixes a bug where `ledger_object::get_u64_field_optional()` checks for 4 bytes instead of 8, causing all u64 field reads from cached ledger objects to fail.

### Context of Change

1. **Code Comparison:** The `current_ledger_object::get_u64_field_optional()` correctly expects 8 bytes
2. **Other Analysis:** All other field getters (u16, u32, UInt128, Hash256) expect the correct byte count
3. **XRPL Spec:** UInt64 fields serialize to 8 bytes per XRPL binary format specification
4. **Rust Spec:** `u64` is 8 bytes on all platforms

### Type of Change

- [ ] Documentation Updates
- [ ] Release

Need to test the change but looking for smart-escrow-rippled.cfg config file to start local rippled. 
CI Tests can help.

### DIFF
--- a/xrpl-wasm-std/src/core/ledger_objects/mod.rs
+++ b/xrpl-wasm-std/src/core/ledger_objects/mod.rs
@@ -460,7 +460,7 @@ pub mod ledger_object {
 
         let result_code = unsafe { get_ledger_obj_field(register_num, field_code, value_ptr, 8) };
 
-        match_result_code_with_expected_bytes_optional(result_code, 4, || Some(value))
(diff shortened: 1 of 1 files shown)


mvadari (approved):

mvadari:
@tekvyy can you re-push this commit after signing? Usually the easiest way is to follow [these instructions](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and then rebase.

PR #38, opened 2025-10-22
Add unit tests

sappenin:
* Tests for blob
* Test for ledger_objects and current ledger_objects


mvadari:
This looks like it's going to conflict heavily with #40 - thoughts on how we can be a bit proactive on aligning them?

PR #52, opened 2025-10-27
Rename the project from `xrpl-wasm-std` to `xrpl-wasm-stdlib`

### LINKED ISSUE #48: Rename project
Let's rename the project to `xrpl-wasm-stdlib`

sappenin:
This PR renames the project from `xrpl-wasm-std` to `xrpl-wasm-stdlib`. If this PR is accepted, then we should also rename the project in Github (old git links will still continue to function).

Fixes #48


mvadari (commented):
change PR name, then this is good to go

PR #73, opened 2025-11-03
Fixes #62 (Update Blob for generic sizes)

### LINKED ISSUE #62: Update Blob to accept arbitrary lengths
The `Blob` struct is generally 1024 bytes long, but in some circumstances we know that we want to store less data (e.g., an NFT URI can only be 256 bytes). Therefore, we should make the Blob struct generic over its data size. 

Once this is implemented, we should refactor [this code in the NFT struct](https://github.com/ripple/xrpl-wasm-std/pull/45/files#diff-f41ffa907aefeab6de59b2f34706f6b693da18e9f3b0b6a0adfc9e7363035bbbR343) to not copy the NFT URI into a blob (instead, we can return a UriBlob or similar and prevent the double-copy. 

```rust
/// Default blob size for general use (memos, etc.)
pub const DEFAULT_BLOB_SIZE: usize = 1024;

/// Blob size for URI fields
pub const URI_BLOB_SIZE: usize = 256;

/// A variable-length binary data container with a fixed maximum size.
///
/// The `Blob` type is generic over its maximum capacity `N`, allowing you to
/// create blobs of different sizes for different use cases. The actual data
/// length is tracked separately in the `len` field.
///
/// # Type Parameters
///
/// * `N` - The maximum capacity of the blob in bytes
///
/// # Examples
///
/// ```
/// use xrpl_wasm_std::core::types::blob::{Blob, DEFAULT_BLOB_SIZE, URI_BLOB_SIZE};
///
/// // Create a standard 1024-byte blob
/// let large_blob: Blob<DEFAULT_BLOB_SIZE> = Blob::new();
///
/// // Create a smaller 256-byte blob for URIs
/// let uri_blob: Blob<URI_BLOB_SIZE> = Blob::new();
/// ```
#[derive(Debug, Clone, Copy, Eq, PartialEq)]
#[repr(C)]
pub struct Blob<const N: usize> {
    pub data: [u8; N],

    /// The actual length of this blob, if less than data.len()
    pub len: usize,
}

impl<const N: usize> Blob<N> {
    /// Creates a new empty blob with the specified capacity.
    #[inline]
    pub const fn new() -> Self {
        Self {
            data: [0u8; N],
            len: 0,
        }
    }

    /// Creates a blob from a byte slice, copying up to N bytes.
    #[inline]
    pub fn from_slice(slice: &[u8]) -> Self {
        let mut data = [0u8; N];
        let len = slice.len().min(N);
        data[..len].copy_from_slice(&slice[..len]);
        Self { data, len }
    }

    /// Returns the actual length of the data in the blob.
    #[inline]
    pub const fn len(&self) -> usize {
        self.len
    }

    /// Returns true if the blob contains no data.
    #[inline]
    pub const fn is_empty(&self) -> bool {
        self.len == 0
    }

    /// Returns the maximum capacity of the blob.
    #[inline]
    pub const fn capacity(&self) -> usize {
        N
    }

    /// Returns a slice of the actual data (not including unused capacity).
    #[inline]
    pub fn as_slice(&self) -> &[u8] {
        &self.data[..self.len]
    }
}

impl<const N: usize> Default for Blob<N> {
    fn default() -> Self {
        Self::new()
    }
}

/// Type alias for the standard 1024-byte blob (for memos and general use)
pub type StandardBlob = Blob<DEFAULT_BLOB_SIZE>;

/// Type alias for 256-byte blob (for URIs and smaller fields)
pub type UriBlob = Blob<URI_BLOB_SIZE>;

/// Empty blob constant for the standard size
pub const EMPTY_BLOB: StandardBlob = Blob {
    data: [0u8; DEFAULT_BLOB_SIZE],
    len: 0usize,
};
```

sappenin:

* Updates the Blob struct and impl to support generic sizes.
* Introduces the `Signature` struct for type safety
* Updates the NFT impl to avoid a byte copy when loading NFT URI
* Introduces various Blob constant sizes for readability and correctness

### Type of Change

- [X] Refactor (non-breaking change that only restructures code)
- [X] Tests (You added tests for code that already exists, or your new feature included in this PR)


xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -818,10 +811,11 @@ pub mod ledger_object {
             let account = AccountID::get_from_current_ledger_obj(sfield::Account).unwrap();
             assert_eq!(account.0.len(), ACCOUNT_ID_SIZE);
 
-            let blob = Blob::get_from_current_ledger_obj(sfield::PublicKey).unwrap();
+            let blob: Blob<{ PUBLIC_KEY_BUFFER_SIZE }> =
```
mvadari: Why are the brackets needed here?

PR #88, opened 2025-11-04
add more getters, simplify `get_field` code across different versions

mvadari:

This PR:
* Adds some missing getters (like `u8`, `Issue`, `Currency`)
* Unifies the getters to drastically reduce the duplicated code between all of the slight variations on basically the same code

This results in a gas savings and binary savings of 7-8% in the gas benchmark as well.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,6 @@
 name: Test
+permissions:
+  contents: read
 
 on:
   push:
@@ -152,3 +154,29 @@ jobs:
 
       - name: Run end-to-end tests
         run: ./scripts/run-tests.sh
+
+  coverage:
+    runs-on: ubuntu-latest
+    env:
+      CARGO_TERM_COLOR: always
+    steps:
+      - uses: actions/checkout@v5
+      - name: Cache dependencies
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+      - name: Install Rust
+        run: rustup update stable
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+      - name: Generate code coverage
+        run: |
+          cargo llvm-cov --all-features --workspace -- --nocapture
+          cd e2e-tests
+          cargo llvm-cov --all-features --workspace --ignore-filename-regex "e2e-tests/.*" -- --nocapture
--- a/e2e-tests/float_tests/src/lib.rs
+++ b/e2e-tests/float_tests/src/lib.rs
@@ -497,3 +497,29 @@ pub extern "C" fn finish() -> i32 {
 
     1 // <-- Finish the escrow to indicate a successful outcome
 }
+
+#[cfg(test)]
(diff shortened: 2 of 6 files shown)


xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -116,143 +119,91 @@ pub trait FieldGetter: Sized {
...
-            unsafe { get_ledger_obj_field(register_num, field_code, value.as_mut_ptr().cast(), 2) };
-        match_result_code_with_expected_bytes_optional(result_code, 2, || {
-            Some(unsafe { value.assume_init() })
-        })
-    }
+/// - `u8` - 1 byte
+/// - `u16` - 2 bytes
+/// - `u32` - 4 bytes
+/// - `u64` - 8 bytes
+trait FixedSizeFieldType: Sized {
```
sappenin: This is great!!

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -340,59 +283,47 @@ impl FieldGetter for Amount {
...
-                register_num,
-                field_code,
-                buffer.as_mut_ptr().cast(),
-                BUFFER_SIZE,
-            )
-        };
-        match_result_code(result_code, || {
-            Amount::from(unsafe { buffer.assume_init() })
-        })
+        match get_variable_size_field::<BUFFER_SIZE, _>(field_code, |fc, buf, size| unsafe {
```
sappenin: Minor nit: Feels like `BUFFER_SIZE` should be pulled out to something (maybe in Amount?), like `AMOUNT_SIZE`?

See `HASH128_SIZE` for an example.

xrpl-wasm-stdlib/src/core/current_tx/mod.rs
```diff
@@ -335,24 +347,22 @@ impl CurrentTxFieldGetter for PublicKey {
 impl CurrentTxFieldGetter for Blob {
     #[inline]
     fn get_from_current_tx(field_code: i32) -> Result<Self> {
-        let mut buffer = core::mem::MaybeUninit::<[u8; 1024]>::uninit();
-        let result_code = unsafe { get_tx_field(field_code, buffer.as_mut_ptr().cast(), 1024) };
-        match_result_code(result_code, || Blob {
-            data: unsafe { buffer.assume_init() },
-            len: result_code as usize,
-        })
+        match get_variable_size_field::<1024, _>(field_code, |fc, buf, size| unsafe {
```
sappenin: Not all Blobs are the same size. Consider these changes in #62. Unclear if we should push that change first, or this one. Thoughts?

(I suppose it means Blobs will get handled like the unsigned ints above?)

sappenin (approved):
Some minor nits, but not worthy of blocking the PR. Feel free to address if desired.

Only question I'm having is, should we merge #92 first before this one -- #92 validates a bunch of byte-handling that I don't think the library has good coverage for currently.

Put another way, I think if we merge #92 first, we'll have better confidence to know if this PR breaks anything.

Thoughts?

xrpl-wasm-stdlib/src/core/current_tx/mod.rs
mvadari: Let's see if we can change this to a generic in a future PR.

mvadari: Resolved in #62

Issue #128, opened 2026-01-14
Consider removing the `keylet_len` parameter from `cache_ledger_obj` 


sappenin:
Currently, `cache_ledger_object...` functions and the primary `cache_ledger_obj` host function accept a byte length for the keylet buffer. However, we should consider removing this length from the function contract, for the following reasons:

1. Keylets are always 32-bytes long (or 34, if encoded with their 2-byte prefix, but in this library they're 32 bytes). 
2. Everywhere in WASM we always hard-code this length to 32 bytes. Or, we use the length of a pre-created buffer that's always 32.
3. An attacker could lie and say the length is 32 but supply a smaller or larger buffer (or vice versa), so the inclusion of this length value isn't adding any security benefit.
4. In the host, keylets always have a know length, so this value should arguably be ignored anyway, to avoid creating a keylet with invalid length.

The new contract should be:

```rust
pub(super) fn cache_ledger_obj(
            keylet_ptr: *const u8,
            cache_num: i32,
        ) -> i32;
```

This removal would percolate through the wasm-std-lib codebase and remove a bunch of redundant usages of the number `32`.

mvadari:
Are you recommending this just in the stdlib or in rippled too?

I believe we decided against this in rippled to keep all buffers and byte arrays consistent - so every pointer needs a length associated with it.

sappenin:
> Are you recommending this just in the stdlib or in rippled too?

Great question -- to be fair, I was mostly thinking about making this library less redundant, but yeah, if we change a host function then that would of course be a change in rippled.

> I believe we decided against this in rippled to keep all buffers and byte arrays consistent - so every pointer needs a length associated with it.

Yeah, I'm remembering that now. I'm just thinking that from a safety perspective, if we allow a WASM caller to specify lengths when they're always 32 bytes, then we have to bounds check those calls, and it just adds one more place that we might get things wrong (in rippled). I need to go look at the C++ implementation in rippled to see if we have a single bounds-check that's well tested and used everywhere, or if we repeat those checks and may miss them sometimes. 

If that code is well written and tested, then maybe this issue is just something to consider for helper functions in this stdlib library (and we leave the host function as out of scope).

mvadari:
@adxdits if you're interested in helping out, we can help direct you to what issues are easiest to pick up/most impactful.

PR #143, opened 2026-02-24
Remove keylet_len parameter from cache_ledger_obj

adxdits:
### Type of Change

- [ ] Documentation Updates
- [ ] Release

fix issue #128


sappenin:
@adxdits Thanks so much for your submission!

See my comment in https://github.com/ripple/xrpl-wasm-stdlib/issues/128#issuecomment-3955419360 for why that issue is no longer valid. Because of that I'm going to close this PR (but feel free to make more comments if you disagree).

PR #148, opened 2026-03-09
fix: Update SFields and SField Ergonomics

### LINKED ISSUE #106: Remove public_key.rs and replace with a PublicKeyBlob typedef in a separate PR
Once #100 merges, then we should remove `public_key.rs` and replace with a `PublicKeyBlob` typedef in `blob.rs`.

sappenin:
This change updates the `sfield.rs` to use the correct types for each sfield. In addition, wrapper Trait APIs are updated to then accept typed `SField` as input. This removes the need to call `.into()` every time a developer wants to call a Trait function using an sField. 

This PR fixes #106

**NOTE**

This PR is part of a series of PRs that decompose #119 (at the request of @mvadari). This stack consists of the
following sub-PRs, whic are expected to be merged into `main` starting with the first in the following list:

1. Fix sfield.rs to relax trait bounds to support `LedgerObjectFieldGetter` or `CurrentTxFieldGetter` (#153)
2. Improve condition/fulfillment parsing (#154)
3. Add unit tests for `current_tx` traits (#155)
4. Add unit tests for `ledger_objects` traits (#156)
5. Fix UB behavior (#194)
6. Update SFields and SField Ergonomics (This PR)


xrpl-wasm-stdlib/src/core/ledger_objects/traits.rs
pwang200, line 209: FYI: sfield::Data is a StandardBlob with size 1024, but XRPL_CONTRACT_DATA_SIZE is 4096. 

I can fit it in a separate PR. WasmBlob size might need to change too.

xrpl-wasm-stdlib/src/sfield.rs
```diff
@@ -338,7 +340,7 @@ pub const PreviousFields: SField<Object, 917510> = SField::new();
 pub const FinalFields: SField<Object, 917511> = SField::new();
 pub const NewFields: SField<Object, 917512> = SField::new();
 pub const TemplateEntry: SField<Object, 917513> = SField::new();
-pub const Memo: SField<Object, 917514> = SField::new();
+pub const Memo: SField<MemoBlob, 917514> = SField::new();
```
pwang200: https://xrpl.org/docs/references/protocol/transactions/common-fields#memos-field
Any reason Memo is changed from Object to MemoBlob?

xrpl-wasm-stdlib/src/sfield.rs
```diff
@@ -259,13 +261,13 @@ pub const Price: SField<Amount, 393244> = SField::new();
...
-pub const MessageKey: SField<StandardBlob, 458754> = SField::new();
-pub const SigningPubKey: SField<StandardBlob, 458755> = SField::new();
-pub const TxnSignature: SField<StandardBlob, 458756> = SField::new();
-pub const URI: SField<StandardBlob, 458757> = SField::new();
+pub const PublicKey: SField<PublicKeyBlob, 458753> = SField::new();
+pub const MessageKey: SField<PublicKeyBlob, 458754> = SField::new();
+pub const SigningPubKey: SField<PublicKeyBlob, 458755> = SField::new();
+pub const TxnSignature: SField<SignatureBlob, 458756> = SField::new();
+pub const URI: SField<UriBlob, 458757> = SField::new();
 pub const Signature: SField<StandardBlob, 458758> = SField::new();
```
pwang200: should Signature be changed as well?

xrpl-wasm-stdlib/src/sfield.rs
sappenin: Great question - worth double-checking with @mvadari on this one since she build the `Object` fallback. But from my perspective, Memos can never be more than 1024 bytes, so defining this as a `MemoBlob` helps when someone needs to allocate enough bytes to hold a memo.

That said, I think the more common way to get at memos will be through locators, so it's possible that no contract author will be allocating bytes for "all memos."

xrpl-wasm-stdlib/src/sfield.rs
sappenin: Yes, good catch. Updated via 943f7702145cc0124451379519ea5882d3bc140b

xrpl-wasm-stdlib/src/core/ledger_objects/traits.rs
sappenin: Good catch. We need to decide what todo here. There's also a TODO [here](https://github.com/ripple/xrpl-wasm-stdlib/blob/main/xrpl-wasm-stdlib/src/core/types/contract_data.rs#L1) in the code.

On the one hand, contracts can currently be 100kb, but this value might change, in which case contracts could break if they're accessing this field.

I propose we do one of these two:

1. Only return a contract fingerprint (e.g., a Sha512-half of the contract bytes). I can't see a reason why anyone would want to load thousands of bytes into the contract itself (outside of checking the contract fingerprint potentially).
2. We disallow getting the contract itself. Then we don't have to worry about how to allocate memory for this case.
3. We allow some kind of paging through this data. [This comment](https://github.com/ripple/xrpl-wasm-stdlib/blob/main/xrpl-wasm-stdlib/src/core/types/contract_data.rs#L19) alludes to this idea, but this kind of thing is the inverse of what that `len` field was designed for (i.e., `len` was meant to allow for a buffer to be big, but "real data" to be less). In addition, the host function interface doesn't allow for any kind of pageability in its current form.

Thoughts?

xrpl-wasm-stdlib/src/sfield.rs
pwang200: thanks

xrpl-wasm-stdlib/src/sfield.rs
sappenin: After discussing, we agreed that `Memo` should use the `Object` type.

sappenin: Reverted via b362366c4af916df23f5ff9efa498adf1e10e809

xrpl-wasm-stdlib/src/core/ledger_objects/traits.rs
sappenin: @pwang200 I created https://github.com/ripple/xrpl-wasm-stdlib/issues/169 for tracking this issue. I agree that a separate PR is the correct place to make this change. 

Resolving this for now (if there's debate to be had around size, we can do that in https://github.com/ripple/xrpl-wasm-stdlib/issues/169).

xrpl-wasm-stdlib/src/core/types/blob.rs
```diff
@@ -20,6 +21,10 @@ pub const FULFILLMENT_BLOB_SIZE: usize = 256;
 // Declared here because there is no Memo struct.
 pub const MEMO_BLOB_SIZE: usize = DEFAULT_BLOB_SIZE;
 
+/// The number of bytes in a Public key. In XRPL, ed25519 public keys are prefixed with a one-byte prefix (i.e., `0xED`)
+/// to be consistent with secp256k1 public keys, which always have 33 bytes.
+pub const PUBLIC_KEY_BLOB_SIZE: usize = 33;
```
pwang200, line 23: Claude flagged that we already had `pub const PUBLIC_KEY_BUFFER_SIZE: usize = 33;` in xrpl-wasm-stdlib/src/core/types/public_key.rs

Do we need both?

pwang200 (approved):
I left a comment, about defining both public key blob size and buffer size. Up to you to keep both or remove one.

xrpl-wasm-stdlib/src/core/types/blob.rs
sappenin: That will be fixed by https://github.com/ripple/xrpl-wasm-stdlib/issues/106

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -384,453 +390,174 @@ pub mod current_ledger_object {
...
-            );
-            assert!(result.is_ok());
-            assert!(result.unwrap().is_some());
+            let account = AccountID::get_from_current_ledger_obj(sfield::Account).unwrap();
+            assert_eq!(account.0.len(), ACCOUNT_ID_SIZE);
 
-            let result = Issue::get_from_ledger_obj_optional(slot, sfield::Asset.into());
-            assert!(result.is_ok());
-            assert!(result.unwrap().is_some());
+            let blob: Blob<PUBLIC_KEY_BUFFER_SIZE> =
```
ricky122-5, line 646: Claude flagged this as a naming inconsistency as PublicKeyBlob can be used, I believe was brought up by Peng earlier and addressed but just highlighting it as it was brought in by the main pull.

ricky122-5 (approved):

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
sappenin: This will be fixed by https://github.com/ripple/xrpl-wasm-stdlib/issues/106. Feel free to leave a comment over there if you think it won't get addressed by new PR work.

pwang200 (approved):

xrpl-wasm-stdlib/src/core/types/blob.rs
```diff
@@ -446,13 +461,6 @@ mod tests {
         assert_eq!(blob2.capacity(), 10);
     }
 
-    #[test]
-    fn test_memo_blob_type_alias() {
```
mvadari, line 450: Why delete this test?

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -9,55 +9,17 @@ use crate::host::error_codes::{
     match_result_code_with_expected_bytes, match_result_code_with_expected_bytes_optional,
 };
 use crate::host::{Result, get_current_ledger_obj_field, get_ledger_obj_field};
+use crate::sfield::SField;
 
 /// Trait for types that can be retrieved from ledger object fields.
 ///
-/// This trait provides a unified interface for retrieving typed data from XRPL ledger objects,
```
mvadari, line 15: Why remove all this?

sappenin: IDK, it didn't seem particularly useful. 

1. For supported types, look at trait impls via IDE.
2. For usage patterns, look at examples or other parts of the code.

That said, if you think this is useful, I can put it back. WDYT?

xrpl-wasm-stdlib/src/core/types/blob.rs
sappenin: The type `MemoBlob` was removed (see above, [here](https://github.com/ripple/xrpl-wasm-stdlib/pull/148/changes#diff-e96c6701f8d04ffa44fc40973ba0cf00c314b715248e731a800c765278129db8L138)) because we decided that `Memo` should be an `Object`. So this test won't compile anymore.

This makes me realize I need to remove `MEMO_BLOB_SIZE` because nothing is using it. Done via b4771e5be90e5a6780088469db9b97e47d974768

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
mvadari: Just didn't seem relevant to this PR, no real objection I guess

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -384,453 +390,174 @@ pub mod current_ledger_object {
     /// * `Err(Error)` - If the field cannot be retrieved or has unexpected size
     #[inline]
     pub fn get_field_optional<T: LedgerObjectFieldGetter, const CODE: i32>(
-        _field: SField<T, CODE>,
-    ) -> Result<Option<T>> {
-        T::get_from_current_ledger_obj_optional(CODE)
-    }
-}
-
-pub mod ledger_object {
```
mvadari, line 393: What happened to this?

xrpl-wasm-stdlib/src/core/current_tx/mod.rs
```diff
@@ -396,4 +421,91 @@ mod tests {
 
         assert!(get_field::<u32, _>(sfield::Flags).is_err());
     }
+
+    // Value-level tests: verify Amount variant detection by populating
+    // the mock buffer with known flag bits + payload.
+
+    #[test]
+    fn test_amount_decodes_xrp_variant() {
```
mvadari: Shouldn't these be in amount.rs?

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -384,453 +390,174 @@ pub mod current_ledger_object {
...
-
-        /// Helper to set up a mock expectation for get_current_ledger_obj_field.
-        ///
-        /// Zero-fills the output buffer before returning. This is required because
-        /// `get_variable_size_field` and the fixed-size getters allocate the buffer
-        /// via `MaybeUninit` and call `assume_init` after the host call returns —
-        /// leaving the buffer uninitialized would be UB.
-        fn expect_current_field(
+        fn expect_current_field<
+            T: LedgerObjectFieldGetter + Send + std::fmt::Debug + PartialEq + 'static,
```
mvadari, line 466: Why is this template so complicated?

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -384,453 +390,174 @@ pub mod current_ledger_object {
...
-        };
+        use crate::core::types::uint::{HASH128_SIZE, HASH256_SIZE, Hash128, Hash256};
         use crate::host::host_bindings_trait::MockHostBindings;
         use crate::host::setup_mock;
-        use crate::sfield;
+        use crate::sfield::{self, SField};
         use mockall::predicate::{always, eq};
 
-        // ========================================
-        // Test helper functions
```
mvadari, line 470: Seems like you're removing a bunch of stuff here, why?

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
sappenin: Short answer: Rust!

Slightly longer answer: it's needed to make mockall work.

(To be fair, I haven't played around with trying to remove these, but will after I address comments)

xrpl-wasm-stdlib/src/core/current_tx/mod.rs
sappenin: Not really. The intent here was to verify the `CurrentTxFieldGetter` impl using something that calls `get_variable_size_field`. Amount served that purpose. 

That said, one variant is sufficient for this, so I removed the other two variants for Amount (and added a comment to reflect why). 

See cd7998f8d88ca987f767c8100d7a504e65aec96a

tools/generateSFields.js
```diff
@@ -116,8 +117,16 @@ async function main() {
...
+    Fulfillment: "FulfillmentBlob",
+    FinishFunction: "WasmBlob",
+    PublicKey: "PublicKeyBlob",
+    Memo: "MemoBlob",
+    Domain: "UriBlob",
+    MessageKey: "PublicKeyBlob",
+    SigningPubKey: "PublicKeyBlob",
+    TxnSignature: "SignatureBlob",
+    URI: "UriBlob",
   }
```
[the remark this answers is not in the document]
sappenin: Fixed via 6635bf7d8c940ac57b634d66d02b820b93292ef2

xrpl-wasm-stdlib/src/core/current_tx/traits.rs
```diff
@@ -230,17 +230,24 @@ pub trait TransactionCommonFields {
...
-        get_field(sfield::SigningPubKey)
+    fn get_signing_pub_key(&self) -> Result<Option<PublicKey>> {
+        get_field(sfield::SigningPubKey).and_then(|blob| match blob.len {
+            0 => Result::Ok(None), // Multi-signature transaction
+            33 => Result::Ok(Some(PublicKey::from(blob.data))), // Single-signature transaction
+            _ => Result::Err(crate::host::Error::from_code(
+                crate::host::error_codes::INTERNAL_ERROR,
+            )),
+        })
     }
```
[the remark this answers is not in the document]
sappenin: This is fixed in `field_helpers.rs` instead, per ebe4efde909cdf195f46193a2f43de3373ac693f

xrpl-wasm-stdlib/src/core/current_tx/traits.rs
```diff
@@ -1119,8 +1126,11 @@ mod tests {
                 let seq_result = tx.get_sequence();
                 assert!(seq_result.is_err());
 
+                // SigningPubKey is special: zero length indicates multi-signature transaction
+                // and should return Ok(None), not an error
                 let signing_key_result = tx.get_signing_pub_key();
-                assert!(signing_key_result.is_err());
+                assert!(signing_key_result.is_ok());
+                assert!(signing_key_result.unwrap().is_none());
```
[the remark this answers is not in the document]
sappenin: This concern is invalid. Tests compile and pass as-is. SField derives PartialEq for same-type comparison and implements Borrow<i32>, which is how mockall's eq() predicate resolves the match against the incoming i32 field code — no PartialEq<i32> impl needed or warranted. Adding one would blur the type boundary without benefit.

xrpl-wasm-stdlib/src/core/ledger_objects/traits.rs
```diff
@@ -776,9 +781,9 @@ mod tests {
                 .with(eq(1), eq(sfield::BurnedNFTokens), always(), eq(4))
                 .times(1)
                 .returning(|_, _, _, _| FIELD_NOT_FOUND);
-            // domain - variable size field, returns 0 for empty (Some with len=0) - StandardBlob uses 1024 bytes
+            // domain - variable size field, returns 0 for empty (Some with len=0)
             mock.expect_get_ledger_obj_field()
-                .with(eq(1), eq(sfield::Domain), always(), eq(1024))
+                .with(eq(1), eq(sfield::Domain), always(), eq(DOMAIN_BLOB_SIZE))
                 .times(1)
```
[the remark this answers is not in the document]
sappenin: This concern is invalid. Mock expectations here pass SField values into eq(...), but the mocked get_ledger_obj_field signature takes the field code as i32. These expectations will not type-check unless the argument is converted (e.g. i32::from(sfield::Domain)). Please apply the same fix to other eq(sfield::...) occurrences in this file's mocks.

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -384,453 +390,174 @@ pub mod current_ledger_object {
...
-            mock.expect_get_ledger_obj_field()
-                .with(eq(slot), eq(field_code), always(), eq(size))
+                .with(eq(field), always(), eq(size))
                 .times(times)
-                .returning(move |_, _, buf, buf_size| {
-                    unsafe { core::ptr::write_bytes(buf, 0, buf_size) };
-                    size as i32
-                });
+                .returning(move |_, _, _| size as i32);
         }
```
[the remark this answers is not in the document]
sappenin: This is no longer an issue due to ebe4efde909cdf195f46193a2f43de3373ac693f, which removes the UB in `field_helpers.rs`

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -384,453 +390,174 @@ pub mod current_ledger_object {
...
+            field: SField<T, CODE>,
             buf_size: usize,
             returned: i32,
         ) {
             mock.expect_get_current_ledger_obj_field()
-                .with(eq(field_code), always(), eq(buf_size))
+                .with(eq(field), always(), eq(buf_size))
                 .times(1)
                 .returning(move |_, _, _| returned);
         }
```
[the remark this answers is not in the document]
sappenin: This is no longer an issue due to ebe4efde909cdf195f46193a2f43de3373ac693f, which removes the UB in `field_helpers.rs`

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -975,188 +688,309 @@ pub mod ledger_object {
...
+            times: usize,
+        ) {
+            mock.expect_get_ledger_obj_field()
+                .with(eq(slot), eq(field), always(), eq(size))
+                .times(times)
+                .returning(move |_, _, buf, buf_size| {
+                    unsafe { core::ptr::write_bytes(buf, 0, buf_size) };
+                    size as i32
+                });
+        }
```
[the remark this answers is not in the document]
sappenin: This is no longer an issue due to ebe4efde909cdf195f46193a2f43de3373ac693f, which removes the UB in `field_helpers.rs`

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -975,188 +688,309 @@ pub mod ledger_object {
...
+            slot: i32,
+            field: SField<T, CODE>,
+            buf_size: usize,
+            returned: i32,
+        ) {
+            mock.expect_get_ledger_obj_field()
+                .with(eq(slot), eq(field), always(), eq(buf_size))
+                .times(1)
+                .returning(move |_, _, _, _| returned);
         }
```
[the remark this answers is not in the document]
sappenin: This is no longer an issue due to ebe4efde909cdf195f46193a2f43de3373ac693f, which removes the UB in `field_helpers.rs`

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
sappenin: It was moved to line 694. But some of the comments were removed by Claude. I'll look through them back in a different commit.

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
sappenin: Errant. Restored via 886dbbfabbadf90cbbc7d8fa003dc373c8206679

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
mvadari: What do you mean moved to line 694? That's deep in the tests, and this is source code

sappenin: Yeah, looks like some line numbers changed since my comment. I'm now seeing line 816 of the proposed merge code (each of the modules has some tests embedded, FWIW).

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
sappenin: Ok, note first this is a test-only method. These trait bounds are imposed by mockall (not by our own trait hierarchy) and without them errors would surface as cryptic compiler errors. Pulling them up to the function signature makes the constraint explicit and keeps the test bodies clean.

- PartialEq — required by mockall::predicate::eq(field), which must compare the captured value against what actually gets passed to the mock.
- std::fmt::Debug — required by mockall to format the predicate in assertion failure messages.
- Send + 'static — required by mockall's .returning(...) closure, which stores the captured field value in the mock's internal state; that storage must be Send + 'static to be thread-safe and heap-storable.

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
sappenin: Would it be helpful if I pulled these test changes into a different PR? This PR does add and adjust test coverage to be more correct, but it has the downside of moving the `ledger_object` much further down which makes it more difficult to review.

mvadari (approved):

PR #164, opened 2026-04-08
more types unit tests

pwang200:


xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -544,6 +580,12 @@ pub mod ledger_object {
 
             expect_current_field(&mut mock, sfield::Flags.into(), 4, 1);
             expect_current_field(&mut mock, sfield::Account.into(), ACCOUNT_ID_SIZE, 1);
+            expect_current_field(&mut mock, sfield::Amount.into(), 48, 1);
```
ricky122-5: kind of a nitpick but if you could ask claude to fix magic numbers would make this more robust to change

xrpl-wasm-stdlib/src/core/current_tx/mod.rs
```diff
@@ -256,3 +256,131 @@ pub fn get_field_optional<T: CurrentTxFieldGetter, F: Into<i32>>(
...
+
+        let _guard = setup_mock(mock);
+
+        assert!(get_field::<u32, _>(sfield::Flags).is_ok());
+        assert!(get_field::<AccountID, _>(sfield::Account).is_ok());
+
+        let result = get_field_optional::<u32, _>(sfield::Flags);
+        assert!(result.is_ok());
+        assert!(result.unwrap().is_some());
+    }
```
ricky122-5, line 385: can we add a error path test? 
from claude
```rust
#[test]
fn test_get_field_returns_err_on_host_error() {
    let mut mock = MockHostBindings::new();
    mock.expect_get_tx_field()
        .with(eq::<i32>(sfield::Flags.into()), always(), eq(4))
        .times(1)
        .returning(|_, _, _| crate::host::error_codes::INTERNAL_ERROR);

    assert!(get_field::<u32, _>(sfield::Flags).is_err());
}
```

ricky122-5 (changes requested):

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
pwang200: thanks, changed to AMOUNT_SIZE

xrpl-wasm-stdlib/src/core/current_tx/mod.rs
pwang200: thanks, this is a good test, added

ricky122-5 (approved):
looks good! thanks!

PR #179, opened 2026-05-29
docs: Update CLAUDE.md with standard library guidance

ricky122-5:

Adds CLAUDE.md for agents

### Type of Change

- [x] Documentation Updates
- [ ] Release


CLAUDE.md
```diff
@@ -0,0 +1,145 @@
...
+- Anything new added to `HostBindings` must be implemented in all three files. CI's `host-function-audit.sh` compares the trait against rippled's exports — keep them in sync.
+
+## Architecture: layering inside `xrpl-wasm-stdlib`
+
+```
+src/
+├── lib.rs            # no_std toggle, panic_handler (wasm only), hex decode helpers, re-exports r_address!
+├── host/             # Low-level layer: HostBindings trait + 3 impls, error codes, trace, field_helpers
+├── core/             # High-level safe API — what contract authors should call
+│   ├── current_tx/   # EscrowFinish marker + traits → typed access to the current TX's fields
```
sappenin: This is probably fine for now, but might be worth adding something into here like, "after any task, consider whether this file should be updated" so that Claude keeps this up to date?

sappenin: @ricky122-5 Thanks for making this change. For future, do you mind making a small comment when you address changes? Something even simple like, "Fixed" or even "Fixed via c54fa389d51e7248da8ed73faa70320bd899392b"

That will help me notice better that you've addressed my comments.

sappenin (approved):

PR #180, opened 2026-06-01
Float wrappers for conversions

ricky122-5:

Adds wrappers for OpaqueFloat operations on conversion host functions

### Context of Change

Wrapper architecture currently is a simple interface into the host functions.

### Type of Change

- [x] Tests (You added tests for code that already exists, or your new feature included in this PR)

Added one success and one error path test for each new wrapper.

### DIFF
--- a/xrpl-wasm-stdlib/src/core/types/opaque_float.rs
+++ b/xrpl-wasm-stdlib/src/core/types/opaque_float.rs
@@ -50,98 +50,6 @@ pub enum RoundingMode {
 }
 
 impl OpaqueFloat {
-    // /// Accessor for the 8-bit `exponent` component of an `OpaqueFloat`.
-    // ///
-    // /// WARNING: Use with caution. In general, `OpaqueFloat` should not be deconstructed; prefer
-    // /// host functions instead (e.g., for things like math operations).
-    // pub fn get_exponent(&self) -> u8 {
-    //     // In big-endian form, the exponent is:
-    //     // * Last 6 bits from bytes[0] (shifted to make room for the 2 bits from bytes[1])
-    //     // * First 2 bits from bytes[1] (shifted down to the lowest position)
-    //     let exponent: u8 = ((self.0[0] & 0x3F) << 2) | ((self.0[1] & 0xC0) >> 6);
-    //
-    //     exponent
-    // }
-    //
-    // /// Accessor for the 54-bit `mantissa` of an `OpaqueFloat`.
-    // ///
-    // /// WARNING: Use with caution. In general, `OpaqueFloat` should not be deconstructed; prefer
-    // /// host functions instead (e.g., for things like math operations).
-    // pub fn get_mantissa(&self) -> u64 {
-    //     // Extract the 54-bit mantissa from the remaining bytes.
-    //     // The mantissa starts from the last 6 bits of the second byte and continues through the
-    //     // remaining bytes.
-    //
-    //     // Extract the top 6 bits from the second byte (after the 2 exponent bits)
-    //     let top_6_bits = (self[1] & 0x3F) as u64;
-    //
-    //     // Extract the remaining 48 bits from self 2-7
-    //     let next_8_bits = (self[2] as u64) << 40;
-    //     let next_16_bits = (self[3] as u64) << 32;
-    //     let next_24_bits = (self[4] as u64) << 24;
-    //     let next_32_bits = (self[5] as u64) << 16;
-    //     let next_40_bits = (self[6] as u64) << 8;
-    //     let next_48_bits = self[7] as u64;
-    //
-    //     // Combine all the bits to form the 54-bit mantissa
-    //     let mantissa = (top_6_bits << 48)
-    //         | next_8_bits
-    //         | next_16_bits
-    //         | next_24_bits
-    //         | next_32_bits
-    //         | next_40_bits
-    //         | next_48_bits;
-    //
-    //     mantissa
-    // }
-    //
-    //     /// Create a new `OpaqueFloat` from an exponent and mantissa.
-    //     ///
-    //     /// WARNING: Use with caution. In general, `OpaqueFloat` should be constructed using
-    //     /// host functions instead of manually setting the exponent and mantissa.
-    //     ///
-    //     /// # Arguments
-    //     ///
-    //     /// * `exponent` - An 8-bit exponent value
-    //     /// * `mantissa` - A 54-bit mantissa value
-    //     ///
-    //     /// # Panics
-    //     ///
-    //     /// Panics if the mantissa exceeds 54 bits.
-    //     pub fn from_exponent_and_mantissa(exponent: u8, mantissa: u64) -> Self {
-    //         // Ensure the mantissa fits in 54 bits
-    //         if mantissa > 0x003FFFFFFFFFFFFF {
-    //             panic!("Mantissa exceeds 54 bits");
-    //         }
-    //
-    //         // Create a buffer for the combined value
-    //         let mut bytes = [0u8; 8];
-    //
-    //         // Store the first 6 bits of the exponent in the first byte
-    //         bytes[0] = (exponent >> 2) & 0x3F;
-    //
-    //         // Store the last 2 bits of the exponent in the first 2 bits of the second byte
-    //         bytes[1] = (exponent & 0x03) << 6;
-    //
-    //         // We need to shift the mantissa to align with our byte layout
-    //         // The mantissa is 54 bits, which is 6 bits in the first byte and 48 bits in the remaining 6 bytes
-    //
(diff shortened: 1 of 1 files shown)


xrpl-wasm-stdlib/src/core/types/opaque_float.rs
```diff
@@ -138,6 +155,95 @@ impl OpaqueFloat {
...
+        let rescode = unsafe {
+            float_to_int(
+                self.0.as_ptr(),
+                core::mem::size_of::<OpaqueFloat>(),
+                int_bytes.as_mut_ptr(),
+                8,
+                mode as i32,
+            )
+        };
+        match_result_code_with_expected_bytes(rescode, 8, || i64::from_be_bytes(int_bytes))
```
pwang200: Claude flagged this as a bug, please see if it makes sense to you:
Host writes the int64 little-endian (adjustWasmEndianess is a no-op on LE hosts in rippled's HostFuncWrapper.cpp), so this should be i64::from_le_bytes(int_bytes). As written it byte-swaps on real validators; the test only passes because the mock writes to_be_bytes.

ricky122-5: Will fix, thanks. the endianness was confusing bc there was conflicting truths. Need to look into this with both repos in context

pwang200:
It seems we have a code version consistency issue. The PR is internally inconsistent: it wraps a host function that only exists in the 12-byte float world, yet keeps OpaqueFloat as [u8; 8]. I am also not confident that the docker image and devnet use the same float version. In addition, the [spec change](https://github.com/XRPLF/XRPL-Standards/pull/504) is not merged yet. I think we should sort out the versions first.

xrpl-wasm-stdlib/src/core/types/opaque_float.rs
pwang200: agree, it is confusing. 

look at both repos is the right approach.

ricky122-5:
> It seems we have a code version consistency issue. The PR is internally inconsistent: it wraps a host function that only exists in the 12-byte float world, yet keeps OpaqueFloat as [u8; 8]. I am also not confident that the docker image and devnet use the same float version. In addition, the [spec change](https://github.com/XRPLF/XRPL-Standards/pull/504) is not merged yet. I think we should sort out the versions first.

Saw the byte inconsistency, will fix. I saw that se/supported supports the 12 byte float so going to amend OpaqueFloat in my PR. I believe the versions are inconsistent.

Docker image maps to se/supported on rippled
Devnet on confluence says latest released beta. Not sure if se/supported counts as this.

From claude:
se/supported is NOT the latest beta. git describe puts it at 3.1.0-b0-875-g62d9f40d33 — it's 875 commits ahead of the 3.1.0-b0 tag on an internal Ripple fork. The latest public beta tag is 3.2.0-b7 (on origin/release-3.2), which is a completely separate lineage. The merge base between se/supported and 3.2.0-b7 is way back at ad3d172a1f (a 3.1.x-era fix).

What se/supported actually is: a single-commit branch whose only change is flipping SmartEscrow from Supported::No to Supported::Yes in features.macro. It's the minimal signal that rippled "understands" the SmartEscrow amendment — not a full release.

What the smart escrow devnet runs: the smart-escrow-devnet4 tag (most recent), which lives on the WASM fork of rippled — a separate lineage with WASM execution infrastructure (GasUsed in metadata, hostfunctions, wamr integration, etc.). That fork has been syncing from upstream XRPLF develop, but it's not the same as the public beta release track.

TL;DR:
- Public devnet (devnet.ripple.com) → tracks the mainline release, roughly 3.1.x / 3.2.0-bX
- Smart escrow devnet → smart-escrow-devnet4 (WASM fork, branched from a 3.1.0-b0 era baseline)
- se/supported → just the feature flag flip, used to signal compatibility on the non-WASM side

xrpl-wasm-stdlib/src/core/types/opaque_float.rs
```diff
@@ -138,22 +142,308 @@ impl OpaqueFloat {
...
-    fn from(value: [u8; 8]) -> Self {
+impl From<[u8; 12]> for OpaqueFloat {
+    fn from(value: [u8; 12]) -> Self {
         OpaqueFloat(value)
     }
 }
 
-/// The number `1` in XRPL's custom float format.
-pub const FLOAT_ONE: [u8; 8] = [0xD4, 0x83, 0x8D, 0x7E, 0xA4, 0xC6, 0x80, 0x00];
+/// The number `1` in STNumber format (mantissa=1000000000000000, exponent=-15).
```
pwang200: The byte values here are correct and match rippled's wasm_fixtures/float_tests fixture, but the doc comments mislabel the mantissa/exponent.
  
0x0DE0B6B3A7640000 = 10^18 and 0xFFFFFFEE = -18, not 10^15 / -15. (Both products equal 1.0, so the value is fine — just the annotation is off.) rippled's own comment says the same:

  // FLOAT_ONE: mantissa=0x0DE0B6B3A7640000 (10^18), exponent=0xFFFFFFEE (-18)

  Suggest matching it:

  /// The number `1` in STNumber format (mantissa = 10^18, exponent = -18).
  pub const FLOAT_ONE: [u8; 12] = [
      0x0D, 0xE0, 0xB6, 0xB3, 0xA7, 0x64, 0x00, 0x00, 0xFF, 0xFF, 0xFF, 0xEE,
  ];

  /// The number `-1` in STNumber format (mantissa = -10^18, exponent = -18).
  pub const FLOAT_NEGATIVE_ONE: [u8; 12] = [
      0xF2, 0x1F, 0x49, 0x4C, 0x58, 0x9C, 0x00, 0x00, 0xFF, 0xFF, 0xFF, 0xEE,
  ];

pwang200: Can we drop the large commented-out blocks while we're here? The get_exponent/get_mantissa/from_exponent_and_mantissa methods (~L54–144) and their commented tests (~L447–559) describe the old 8-byte bit-packed format and don't apply to the 12-byte STNumber layout — they're stale and a bit misleading now. (The short // from_int / // to_int section markers are fine to keep.)

ricky122-5: Fixed

ricky122-5: Fixed

xrpl-wasm-stdlib/src/core/types/opaque_float.rs
```diff
@@ -37,239 +34,321 @@
 /// ```no_run
 /// # use xrpl_wasm_stdlib::core::types::opaque_float::OpaqueFloat;
 /// // Create from host function
-/// let mut float_bytes = [0u8; 8];
-/// // float_from_int(100, float_bytes.as_mut_ptr(), 8, 0);
+/// let mut float_bytes = [0u8; 12];
+/// // float_from_int(100, float_bytes.as_mut_ptr(), 12, 0);
```
sappenin: Consider uncommenting `float_from_int`?
```suggestion
/// float_from_int(100, float_bytes.as_mut_ptr(), 12, 0);
```

xrpl-wasm-stdlib/src/core/types/opaque_float.rs
```diff
@@ -37,239 +34,321 @@
...
-    //         bytes[5] = ((mantissa >> 16) & 0xFF) as u8;
-    //         bytes[6] = ((mantissa >> 8) & 0xFF) as u8;
-    //         bytes[7] = (mantissa & 0xFF) as u8;
-    //
-    //         // Create a new OpaqueFloat from the combined bytes
-    //         // ALWAYS use `from_be_bytes` because that's how things are assembled directly above.
-    //         // This will manifest a u64 which will be correct.
-    //         OpaqueFloat(u64::from_be_bytes(bytes))
-    //     }
+    pub fn from_int(i: i64, mode: RoundingMode) -> Result<OpaqueFloat> {
```
sappenin, line 51: Let's add docs to each function?

sappenin (changes requested):

pwang200 (approved):
👍  Please address comments from @sappenin

sappenin (commented):
Per comments by @pwang200, let's hold this change until [the spec](https://github.com/XRPLF/XRPL-Standards/pull/504) is merged.

ricky122-5:
Superseded by the reworked float/`Number` wrappers, rebuilt fresh on the current `xrpl-common-stdlib` layout: #259 (conversion wrappers) and #260 (operation wrappers). This branch predates the crate reorg (`xrpl-wasm-stdlib` paths, `XFloat`/`OpaqueFloat` naming), so closing as stale. Reopen if still needed.

PR #182, opened 2026-06-01
feat: Ledger obj wrappers

ricky122-5:

Added wrappers + tests for chain state

### Type of Change

- [ ] Documentation Updates
- [ ] Release

Success + Error for each test


pwang200:
@ricky122-5  

The xrpld side use mixed endiannesses. It is a bit surprising, but here are the reasons.
- Since Wasm is little-endian, when returning individual ints etc from host, the host convert the bytes to little-endian. Note that this conversion is no-op on most server CPUs. No reason that the host convert to big endian that wasm side must convert back to little endian. 
- However, if the int is part of a serialized data, and the host returns that whole data, the int byte ordering follows the serialized data format. 

Claude's more detailed assessment

"Yes — there's a deliberate mix.

  The rule the code intends: all multi-byte integers go to wasm in little-endian (wasm-native). adjustWasmEndianess (ParamsHelper.h:233) is a
  byte-swap that's a no-op on little-endian hosts and only swaps on big-endian hosts — so the wasm guest always sees little-endian integers
  regardless of host CPU.
  
  How each return type actually lands in wasm memory:

  - Scalar ints (uint32_t, int64_t, FloatPair) — little-endian. returnResult swaps then memcpy (HostFuncWrapper.cpp:321/336/351).
  - Integer leaf fields via getTxField/getLedgerObjField (UInt16/32/64, Int32/64) — little-endian. getIntBytes swaps then copies
  (HostFuncImplGetter.cpp:18). Not canonical XRPL order.
  - STAmount / STNumber fields — big-endian (canonical XRPL serialization via Serializer, HostFuncImplGetter.cpp:98-100). ⚠️ This is the odd one
  out.
  - AccountID, uint256/hashes, blobs (VL), MPTID — raw network-order bytes (byte strings, not byte-swapped).
  - int32_t results (array lengths, status) — returned as the function's i32 return value, not written to memory; wasm handles natively.

  The inconsistency: a wasm guest reading a UInt32 field gets a little-endian u32, but reading an Amount field gets the big-endian serialized
  blob. Both arrive through the same Bytes path (raw memcpy, no adjustment in the wrapper) — the divergence is entirely inside getAnyFieldData:
  integers get swapped to LE, amounts/numbers get canonical BE."

ricky122-5:
> @ricky122-5
> 
> The xrpld side use mixed endiannesses. It is a bit surprising, but here are the reasons.
> 
> * Since Wasm is little-endian, when returning individual ints etc from host, the host convert the bytes to little-endian. Note that this conversion is no-op on most server CPUs. No reason that the host convert to big endian that wasm side must convert back to little endian.
> * However, if the int is part of a serialized data, and the host returns that whole data, the int byte ordering follows the serialized data format.
> 
> Claude's more detailed assessment
> 
> "Yes — there's a deliberate mix.
> 
> The rule the code intends: all multi-byte integers go to wasm in little-endian (wasm-native). adjustWasmEndianess (ParamsHelper.h:233) is a byte-swap that's a no-op on little-endian hosts and only swaps on big-endian hosts — so the wasm guest always sees little-endian integers regardless of host CPU.
> 
> How each return type actually lands in wasm memory:
> 
> * Scalar ints (uint32_t, int64_t, FloatPair) — little-endian. returnResult swaps then memcpy (HostFuncWrapper.cpp:321/336/351).
> * Integer leaf fields via getTxField/getLedgerObjField (UInt16/32/64, Int32/64) — little-endian. getIntBytes swaps then copies
>   (HostFuncImplGetter.cpp:18). Not canonical XRPL order.
> * STAmount / STNumber fields — big-endian (canonical XRPL serialization via Serializer, HostFuncImplGetter.cpp:98-100). ⚠️ This is the odd one
>   out.
> * AccountID, uint256/hashes, blobs (VL), MPTID — raw network-order bytes (byte strings, not byte-swapped).
> * int32_t results (array lengths, status) — returned as the function's i32 return value, not written to memory; wasm handles natively.
> 
> The inconsistency: a wasm guest reading a UInt32 field gets a little-endian u32, but reading an Amount field gets the big-endian serialized blob. Both arrive through the same Bytes path (raw memcpy, no adjustment in the wrapper) — the divergence is entirely inside getAnyFieldData: integers get swapped to LE, amounts/numbers get canonical BE."

Thanks for the explanation, looked into it and it matches the code. One thing is in chain.rs's helpers, I call from_le_bytes(), but FixedSizeFieldType generic impl uses a raw pointer cast that's already LE. Can make an explicit conversion there if needed.

xrpl-wasm-stdlib/src/core/chain.rs
```diff
@@ -0,0 +1,206 @@
+use crate::host::{
+    Error, Result, amendment_enabled as host_amendment_enabled,
+    error_codes::match_result_code_with_expected_bytes, get_base_fee, get_ledger_sqn,
+    get_parent_ledger_hash, get_parent_ledger_time,
+};
+
+pub fn ledger_sqn() -> Result<u32> {
```
sappenin, line 7: Let's discuss the module location of `ledger_sqn`,  `parent_ledger_time`, and `parent_ledger_hash` functions. These values are not specific to "the chain," but are instead from the last validated ledger. 

This has me wondering if these shouldn't live in a module called `ledger` or `validated_ledger`?

(For `base_fee` and `amendment_enabled` I like placing those in a module called `chain`).

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
```diff
@@ -336,6 +339,15 @@ impl LedgerObjectFieldGetter for Hash192 {
     }
 }
 
+pub fn cache_ledger_obj(keylet: &[u8], cache_num: i32) -> Result<i32> {
```
sappenin: Nice catch on this one, I'm surprised it didn't exist. If we keep this function, it's probably worth updating all of the examples and e2e-tests to use it. If we do that, I suggest pulling that into a separate PR.

sappenin (changes requested):

xrpl-wasm-stdlib/src/core/chain.rs
sappenin: Discussed verbally. We'll leave as-is.

xrpl-wasm-stdlib/src/core/ledger_objects/mod.rs
sappenin: Discuss with @ricky122-5 and he'll likely remove this whole function in-favor of a different approach.

sappenin (approved):

PR #184, opened 2026-06-09
Update READMEs with Docker instructions

sappenin:
Update READMEs to better describe Docker usage.


e2e-tests/trace_escrow_account/README.md
```diff
@@ -25,6 +25,10 @@ cargo build --target wasm32v1-none --release
 
 ```shell
 cd ../..
+if ! docker ps -q --filter ancestor=rippleci/xrpld:ripple--se--supported | grep -q .; then
```
ricky122-5, line 28: I think these commands should be consistent with the ones in examples

The docker commands in examples don’t have the docker ps guard, so if a dev ran one example and then tried to run another it would error and be slightly confusing

ricky122-5 (dismissed):

e2e-tests/trace_escrow_account/README.md
sappenin: Are you advocating to remove the if check here?

ricky122-5: I was thinking have the guard on the READMEs for every example. Sorry, missed this

sappenin:
Closing this as stale. https://github.com/ripple/xrpl-wasm-stdlib/pull/291 is a preferred path anyway, allowing scripts to use Docker by default instead of having to declare this in READMEs.

PR #186, opened 2026-06-17
feat: Escrow crate

ricky122-5:

Creates the escrow crate with EscrowFinishContext object.

### Context of Change

https://docs.google.com/document/d/1-waCIzHU_pWNNh2kBYRUCOq9puwBgN-QL_FXBr30X7g/edit?usp=sharing

### Type of Change

- [ ] Documentation Updates
- [ ] Release


xrpl-escrow-stdlib/README.md
```diff
@@ -0,0 +1,34 @@
...
+single entry point a Smart Escrow author interacts with, along with safe wrappers around the
+escrow-unique host functions (e.g., `update_data`). All unsafe FFI is contained here; user code
+stays fully safe.
+
+## Usage
+
+```rust,ignore
+use xrpl_escrow_stdlib::*;
+
+fn run(ctx: EscrowFinishContext) -> Result<bool, ()> {
```
sappenin: I'm not sure that this is the correct function contract (but feel free to push back).

The current `Result<bool, ()>` doesn't give WASM contract developers enough control over the return value. Instead, we need to carry a numeric code in both the success and rejection cases (note that `reject_codes` are distinct from raw XRPL `tec`/`tes` codes).

Take a look at this counter-proposal and let's discuss tomorrow?
https://gist.github.com/sappenin/9be353e1c9d8b1aab0ec12abe2cc67c2

ricky122-5: TBH this is on me, should've elaborated on the docs, but `Result<bool, ()>` isn't the only supported type. The macro also supports `i32` & empty `Result` types.

However, I do agree that it doesn't give contract devs enough control. I tried to solve this by attempting to use `Into<i32>` for parsing error codes, but that would require a `Result<bool, E>` signature.

I like the idea of having that single source of truth as an enum. Would make it a lot simpler for devs. I feel like another question here is if we should keep `i32` as a valid return as well if a dev would like it that way, as the macro checks return types by name.

sappenin: > another question here is if we should keep i32 as a valid return as well 

If we can do that, I like it. We should probably model out what we want contract developers to be able to do, then code around that.

sappenin: Some notes from our conversation:

```rust
fn run(ctx: EscrowFinishContext) -> FinishResult {
  if some_condition {
    5
    //FinishResult::success() 
  } else {
    -5 
    //FinishResult::default() 
  }
}

impl Into<i32> for FinishResult { // Maybe use `From` instead
  fn into(value:i32): FinishRestult {
    if(i32 >0){
  
     return FinishResult::Succeed(NonZeroU32::new(27).unwrap())
    }
   else{

   }
  }
}
```

xrpl-escrow-stdlib/src/ctx/escrow_finish.rs
```diff
@@ -0,0 +1,90 @@
...
+/// Entry-point context for a Smart Escrow finish operation.
+///
+/// Provides access to the current [`EscrowFinish`] transaction via [`SmartFeatureContext::tx`]
+/// and to the escrow ledger object via [`escrow`](EscrowFinishContext::escrow). Escrow-unique
+/// host functions (e.g., [`update_data`](EscrowFinishContext::update_data)) are exposed as safe
+/// inherent methods; no `unsafe` code is needed in user crates.
+///
+/// The `#[smart_escrow]` entry-point macro constructs this type via `Default::default()` and
+/// passes it to the user function. Direct construction is not needed in normal usage.
+pub struct EscrowFinishContext {
```
mvadari: I would make these context changes separate from the new crate

Cargo.toml
```diff
@@ -3,6 +3,7 @@ resolver = "2"
 members = [
     "xrpl-wasm-stdlib",
     "xrpl-wasm-stdlib/xrpl-macros",
+    "xrpl-escrow-stdlib",
```
mvadari, line 6: Examples should be using the new crate

ricky122-5: Right now I want to keep it non functional until the macros have been looped in. Can include example updates in that PR so that examples can have #[smart-escrow]

xrpl-escrow-stdlib/README.md
```diff
@@ -0,0 +1,34 @@
+# xrpl-escrow-stdlib
+
+Smart Escrow entry-point context and host-function wrappers for writing XRPL Smart Escrows in Rust.
+
+This crate is part of the `xrpl-wasm-stdlib` workspace. It provides [`EscrowFinishContext`], the
+single entry point a Smart Escrow author interacts with, along with safe wrappers around the
```
sappenin: Let's clarify terms a bit here. The finish function is the "single entry point" for a Smart Escrow author. It's the "door"  the developer walks through. The Context object is their control surface once inside: everything they need to act. So, how about this:

```suggestion
control surface a Smart Escrow author interacts with, along with with safe, scoped access to
```

sappenin (approved):

PR #187, opened 2026-06-17
feat: Make r_address macro typed + macro crate org

ricky122-5:

Takes r_address logic out of lib.rs, casts it to AccountID

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Trybuild tests to verify compilation:
- `fail_not_r_prefix` — non-`r` prefix rejected
- `fail_bad_checksum` — base58 checksum mismatch rejected
- `fail_too_short` — undersized input rejected
- `fail_non_literal` — runtime expression rejected (string literal required)

Run with:
```bash
cargo test -p xrpl-macros --test r_address

Also verified:
- cargo check --workspace clean on both stdlib and examples workspaces
- notary.wasm builds at 891 B (release, wasm32v1-none) — no binary growth
- Notary example still compiles after switching NOTARY_ACCOUNT to AccountID
- Oracle example still compiles after dropping the redundant AccountID(...) wrap

### CODE REVIEW

**Review by mvadari — COMMENTED** [review]
Is the separate macros crate no longer needed now?

File: xrpl-wasm-stdlib/xrpl-macros/tests/r_address/fail_bad_checksum.rs [hunk 1]
```diff
@@ -0,0 +1,7 @@
+use xrpl_macros::r_address;
```
mvadari: Why design your tests like this?

xrpl-wasm-stdlib/xrpl-macros/Cargo.toml
```diff
@@ -23,3 +23,6 @@ features = ["full"]
 
 [dependencies.quote]
 version = "1.0"
+
+[dev-dependencies]
+trybuild = "1.0"
```
mvadari, line 31: What does this dependency do?

xrpl-wasm-stdlib/xrpl-macros/README.md
```diff
@@ -18,21 +18,22 @@ Then use the macro:
...
 
-A compile-time macro for converting XRPL classic addresses (r-addresses) to 20-byte arrays.
+A compile-time macro for converting XRPL classic addresses (r-addresses) to typed `AccountID` values.
 
 ## Features
 
 - **Zero runtime overhead**: Address decoding happens at compile time
-- **Type safe**: Invalid addresses cause compilation errors
-- **No binary bloat**: The final WASM contains only the raw 20-byte array, no decoding logic
+- **Type safe**: Invalid addresses cause compilation errors; the result is a typed `AccountID`, not a raw byte array
```
mvadari:
```suggestion
- **Type safe**: Invalid addresses cause compilation errors; the result is a typed `AccountID`
```
The "not a raw byte array" part is only relevant in the PR context, it won't make any sense 2 months from now

xrpl-wasm-stdlib/xrpl-macros/src/lib.rs
```diff
@@ -1,85 +1,7 @@
-//! The `r_address!` macro for compile-time address conversion converts XRPL classic addresses (r-addresses)
```
mvadari, line 1: Why remove the docs here?

xrpl-wasm-stdlib/xrpl-macros/src/lib.rs
```diff
@@ -1,85 +1,7 @@
...
-    let second = Sha256::digest(first);
-    if &second[0..4] != checksum {
-        return None;
-    }
-    // Payload is version (1) + 20 bytes account id
-    if payload.len() != 1 + 20 {
-        return None;
-    }
-    Some(payload[1..].to_vec())
+    r_address::expand(input)
```
mvadari: What code change allows this to be interpreted as an AccountID now?

xrpl-wasm-stdlib/xrpl-macros/src/lib.rs
sappenin: For some reason Claude is doing this to me a lot lately (i.e., aggressively removing code comments). I think we could benefit from a `CLAUDE.md` that basically says, "don't remove code comments inside of files. Instead, rewrite them to be accurate, concise, and useful" or some such.

xrpl-wasm-stdlib/xrpl-macros/tests/r_address/fail_bad_checksum.rs
sappenin: I'm intrigued by half of this style change.

More specifically, I kinda like the idea of moving tests to their own file (though I don't think that's idiomatic). It can be pretty hard to read Rust code that has like 3 lines of implementation, and then 1000 lines of tests.

That said, putting each test into a separate file is interesting - I haven't seen it before; though I think I would just put this all into one file called `/tests/r_address.rs`.

Definitely worth more discussion here I think.

ricky122-5: The reason for the perfile tests is trybuild introduces compile-fail tests. Each failure essentially has to be its own .rs file so that trybuild can pair it with its stderr output.

xrpl-wasm-stdlib/xrpl-macros/Cargo.toml
ricky122-5: This is the trybuild crate I was talking about in my thread about macros. It's just a compile-time crate that verifies the validation logic for our macro is processing its inputs correctly and flagging invalid entries correctly, it's dev-only

xrpl-wasm-stdlib/xrpl-macros/README.md
ricky122-5: Good call, will fix thanks

xrpl-wasm-stdlib/xrpl-macros/src/lib.rs
ricky122-5: Well actually I did this on purpose lol. Because all the logic for r_address moved into its own standalone file, I just moved them to the code they describe in r_address.rs


Let me know if this should stay. I just didn't want to crowd the file later when I add all my macros.

xrpl-wasm-stdlib/xrpl-macros/src/lib.rs
ricky122-5: The macro emits `::xrpl_wasm_stdlib::core::types::account_id::AccountID(bytes)` instead of the [u8; 20]. AccountID is a #[repr(transparent)] newtype over the [u8;20], so it can be constructed. in r_address:42 it should show that.

xrpl-wasm-stdlib/xrpl-macros/src/lib.rs
sappenin: Oh, my bad - you're right, this documentation belongs in `r_address.rs`, which is already there.

I suppose that prompts another question though: what's the design intent around separating the macro into different files between `lib.rs` and `r_address.rs`?

ricky122-5:
> Is the separate macros crate no longer needed now?

Nope, we still need them. This PR is the first step and just breaking down the macros organization. We're later going to promote it to its own standalone crate (with proc-macro=True) that the escrow/contract crate will need to consume.

xrpl-wasm-stdlib/xrpl-macros/src/lib.rs
ricky122-5: In my design `lib.rs` houses all the proc macro declarations, so in other PRs I will be adding types & entry point declarations. Wanted to separate logic to prevent a messy `lib.rs` file.

xrpl-wasm-stdlib/xrpl-macros/src/lib.rs
mvadari: That seems risky given that the macro is in a separate crate - that suggests that using this crate without the stdlib crate won't work.

ricky122-5:
For everything that doesn't need to rely on compile-time errors, I refactored to remove trybuild tests and make them unit tests. 

One try-build test remains: usage with non-literals.

Trybuild is something that I foresee will get very useful for codegen macros, something which its success metric isn't "did this do this right" but rather "is this code valid", so I feel like even one test can't hurt bringing in the dep, although it is a valid test case.

sappenin (approved):
LGTM!

xrpl-wasm-stdlib/xrpl-macros/src/r_address.rs
```diff
@@ -0,0 +1,154 @@
...
+pub fn expand(input: TokenStream) -> syn::Result<TokenStream> {
+    let addr_lit = syn::parse2::<LitStr>(input)?;
+    let addr = addr_lit.value();
+
+    let bytes = decode_classic_address_to_20bytes(&addr).map_err(|reason| {
+        syn::Error::new(addr_lit.span(), format!("Invalid r-address: {reason}"))
+    })?;
+
+    let bytes_tokens = bytes.iter().map(|b| quote! { #b });
+    let expanded = quote! {
```
mvadari, line 34: Not worth doing here since you've already got approvals, but in the future I'd split up moving the code around (trivial to review) and the functional change, especially since your functional change is buried deep in the moved code and easy to miss.

ricky122-5: Gotcha, will keep in mind from now on. Thank you

PR #189, opened 2026-06-18
feat: Typed Constant Macros

ricky122-5:

Adds macros for typed constants at compile time, with trybuild tests.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Run from repo root:

- `cargo test --workspace` — all targets pass
  - `xrpl-macros` lib: 0 tests (decoder unit tests removed; coverage moved to integration + trybuild)
  - `xrpl-macros` trybuild (`tests/compile_fail.rs`): 19 compile-fail cases across r_address / hash256 / pubkey / currency / blob
  - `xrpl-wasm-stdlib` integration (`tests/macros.rs`): 16 happy-path cases, each macro invoked in a `const` binding and the resulting `AccountID` / `Hash256` / `PublicKey` / `Currency` / `Blob<N>` asserted byte-for-byte
  - existing `xrpl-wasm-stdlib` unit + e2e suites untouched and still pass (279 unit + 38 e2e)
- `cargo build --workspace --release` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo fmt --check` — clean
- Regenerate trybuild snapshots if any error message changes: `TRYBUILD=overwrite cargo test -p xrpl-macros --test compile_fail`

Coverage notes:
- Happy path covers checking that macro compiles into type and contents pass through correctly.
- Compile-fail cases cover non-literal input
- Unit tests cover wrong length, invalid hex chars, and per-macro semantic rejections (XRP for `currency!`, bad EC prefix for `pubkey!`, bad checksum / wrong prefix for `r_address!`, capacity overflow for `blob!`).

## Future Tasks
Break macro crate off into standalone root level crate.


ricky122-5:
This is based off of the r_address branch, meant to be merged after

xrpl-wasm-stdlib/xrpl-macros/src/currency.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    #[test]
+    fn rejects_nonstandard_non_hex() {
+        let key = "GGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGGG";
+        assert!(decode_currency(key).is_none());
+    }
+
+    #[test]
+    fn rejects_wrong_length() {
+        assert!(decode_currency("US").is_none());
```
sappenin: Shouldn't this error instead? Same for the other invalid cases?

xrpl-wasm-stdlib/xrpl-macros/src/currency.rs
```diff
@@ -0,0 +1,118 @@
...
+    #[test]
+    fn rejects_non_alphanumeric() {
+        assert!(decode_currency("U$D").is_none());
+    }
+
+    #[test]
+    fn nonstandard_hex() {
+        let key = "0158415500000000C1F76FF6ECB0BAC600000000";
+        let bytes = decode_currency(key).unwrap();
+        assert_eq!(bytes[0], 0x01);
```
sappenin: Seems like we should assert all bytes in `key` here.

sappenin (changes requested):

xrpl-wasm-stdlib/xrpl-macros/src/currency.rs
ricky122-5: Good catch, thank you, will fix!

xrpl-wasm-stdlib/xrpl-macros/src/currency.rs
ricky122-5: Quick question: do you mean the assert should be is_err()? the fn types are Option which is the convention that r_address used. Thanks

ricky122-5: This should be fixed

xrpl-wasm-stdlib/xrpl-macros/src/currency.rs
ricky122-5: Should be fixed. Thanks!

xrpl-wasm-stdlib/xrpl-macros/src/blob.rs
```diff
@@ -34,9 +34,9 @@ pub fn expand(input: TokenStream) -> TokenStream {
     let hex_str = hex.value();
 
     let bytes = match decode_blob_hex(&hex_str) {
-        Some(b) => b,
-        None => {
-            return syn::Error::new(hex.span(), format!("Invalid blob hex: {hex_str}"))
+        Ok(b) => b,
```
sappenin: At the risk of contradicting (just a bit) what I said earlier today, it might be nice to use a `Result` in the `expand` function, so that we can unit test the error messages? 

Experimenting in my own IDE, it looks like these minor changes might work. Take a look and let me know what you think?

1. In `blob.rs`:

```rust
pub fn expand(input: TokenStream) -> syn::Result<TokenStream> {
    let BlobInput { hex, capacity } = syn::parse::<BlobInput>(input)?;
    let hex_str = hex.value();

    let bytes = decode_blob_hex(&hex_str)
        .map_err(|reason| syn::Error::new(hex.span(), format!("Invalid blob hex: {reason}")))?;

    let n = match &capacity {
        Some(lit) => lit.base10_parse::<usize>()?,
        None => bytes.len(),
    };

    check_capacity(bytes.len(), n)
        .map_err(|msg| syn::Error::new(hex.span(), msg))?;

    let len = bytes.len();
    let mut data = bytes;
    data.resize(n, 0u8);
    let data_tokens = data.iter().map(|b| quote! { #b });

    let expanded = quote! {
        ::xrpl_wasm_stdlib::core::types::blob::Blob::<#n> {
            data: [#(#data_tokens),*],
            len: #len,
        }
    };
    Ok(TokenStream::from(expanded))
}
```

2. In lib.rs:

```rust
#[proc_macro]
pub fn blob(input: TokenStream) -> TokenStream {
    match blob::expand(input) {
        Ok(tokens) => tokens,
        Err(err) => err.to_compile_error().into(),
}
```

3. In unit tests in blob.rs

```rust
 #[test]
    fn rejects_odd_length() {
        let err = decode_blob_hex("ABC").unwrap_err();
        assert_eq!(err, "hex string must have an even number of characters");
    }

    #[test]
    fn rejects_non_hex() {
        let err = decode_blob_hex("ZZ").unwrap_err();
        assert_eq!(err, "non-hex character");
    }
```

or this is if you prefer:

```rust
    #[test]
    fn rejects_odd_length() {
      let result = decode_blob_hex("ABC");
      assert!(result.is_err());
      assert_eq!(result.unwrap_err(), "hex string must have an even number of characters");
    }
```

ricky122-5: Correct me if I'm wrong but aren't those tests just calling the decode functions that have the Result return? I think you can still test the err messages through that

Maybe were you wanting unit tests at the expand level?

sappenin: Oh, yeah - I was focusing on unit testing the `expand` method. For example, imagine your unit tests covering `decode` pass just fine. That doesn't prove anything about the implementation of your `expand` function.

That said, if you assert the error message in the unit tests for decode, _and_ you assert that decode is called properly in your `expand` function, then that would suffice too, but it's more difficult to test because `decode` isn't a mock.

xrpl-wasm-stdlib/src/lib.rs
```diff
@@ -4,7 +4,10 @@
 #[cfg(not(target_arch = "wasm32"))]
 extern crate std;
 
-// Re-export the r_address macro for convenient access
+pub use xrpl_macros::blob;
```
sappenin, line 8:
```suggestion
// Re-export macros for convenient access
pub use xrpl_macros::blob;
```

xrpl-wasm-stdlib/tests/macros.rs
```diff
@@ -22,3 +29,72 @@ fn r_address_expands_to_const_account_id() {
         ]
     );
 }
+
+#[test]
```
sappenin, line 33: Remind me - is the reason these tests are here (in a `/tests` folder as opposed to in `/src/lib.rs` because of `trybuild`?

I don't see any failure cases in there, so it feels like these tests should just be moved over to the implementation.

xrpl-wasm-stdlib/xrpl-macros/src/hex_util.rs
```diff
@@ -0,0 +1,11 @@
+//! Shared hex-decoding helpers used by the typed-constant macros.
+
+/// Decode an even-length hex string into bytes. Caller must verify length and
+/// that every char is `is_ascii_hexdigit` before calling — non-hex input will
+/// panic.
+pub(crate) fn decode_hex(s: &str) -> Vec<u8> {
```
sappenin, line 6: Need some unit tests for this function.

sappenin (changes requested):

xrpl-wasm-stdlib/tests/macros.rs
ricky122-5: The reason they're here is because proc macro crates don't allow you to re export types. So to work with making the macro actually emit the typed constant and verifying it, they have to live somewhere I can bring in the types. I just chose tests/ because it was a natural place, but open to other ideas / suggestions.

sappenin: Ah, yes - my bad. Resolving.

sappenin (approved):

PR #191, opened 2026-06-22
feat: Entry point scaffold

ricky122-5:

Adds the skeleton code for the entry point macros for Smart Escrows and Smart Contracts

### Type of Change

- [ ] Documentation update
- [ ] Release

## Test Plan
For skeleton code, these are passthroughs. Compilation is the test here

## Future Tasks
Implement validation, codegeneration, and full implementation
Add unit / trybuild tests


sappenin (dismissed):

ricky122-5:
Going to merge after #189

xrpl-wasm-stdlib/xrpl-macros/src/lib.rs
```diff
@@ -139,3 +144,39 @@ pub fn blob(input: TokenStream) -> TokenStream {
         Err(err) => err.to_compile_error().into(),
     }
 }
+
+/// Wraps a Smart Escrow finish function in the `extern "C" fn finish()` entry point
+/// the XRPL host calls when an `EscrowFinish` transaction invokes the feature.
+///
+/// The annotated function must:
+/// - Take `EscrowFinishContext` as its first (and only) argument.
+/// - Return `Result<bool, E>`, `Result<(), E>`, or `i32`.
```
sappenin: Is `Result<bool, E>` still a thing?

To ease contract developer mental overhead, I prefer to only give them two options (instead of three). e.g., we should only allow these two in the response:

1. A `FinishResult` (or `SmartFeatureResult` or whatever we call that thingy) --> provides an easy to use guardrail that difficult to mess up.
2. An `i32` --> if a developer wants "very low level access" and knows what they're doing

Introducing an additional `bool` just adds one more thing for a contract developer to understand, but it's not adding any value beyond #1 above, which a developer needs to understand anyway (especially before they reduce their options into a `bool`).

Thoughts?

sappenin (changes requested):

xrpl-wasm-stdlib/xrpl-macros/src/lib.rs
ricky122-5: I agree that these two choices should be the options, will make that change. Thanks!

ricky122-5: Fixed and added two examples for both i32 / FinishResult

sappenin (approved):

PR #194, opened 2026-06-22
fix: Use zeroed buffers for variable-size field helpers to prevent UB

sappenin:

For variable-size fields, result_code == 0 is a valid success (e.g. an empty SigningPubKey signals a multi-signed transaction). Using uninit() meant assume_init() could be called on uninitialized memory in that case; zeroed() eliminates this undefined behavior.

Also annotate the fixed-size helpers to explain why uninit() remains correct there (assume_init is only called when exactly N bytes were written).

### Context of Change

Recommended by AI code review during review of https://github.com/ripple/xrpl-wasm-stdlib/pull/148

### Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/xrpl-wasm-stdlib/src/host/field_helpers.rs
+++ b/xrpl-wasm-stdlib/src/host/field_helpers.rs
@@ -44,6 +44,7 @@ pub fn get_fixed_size_field_with_expected_bytes<const N: usize, F>(
 where
     F: FnOnce(i32, *mut u8, usize) -> i32,
 {
+    // uninit is safe: assume_init is only called when result_code == N, meaning the host wrote exactly N bytes.
     let mut buffer = core::mem::MaybeUninit::<[u8; N]>::uninit();
     let result_code = host_fn(field_code.into(), buffer.as_mut_ptr().cast(), N);
     match_result_code_with_expected_bytes(result_code, N, || unsafe { buffer.assume_init() })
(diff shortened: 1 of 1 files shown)


xrpl-wasm-stdlib/src/host/field_helpers.rs
```diff
@@ -135,11 +138,13 @@ pub fn get_variable_size_field<const N: usize, F>(
 where
     F: FnOnce(i32, *mut u8, usize) -> i32,
 {
-    let mut buffer = core::mem::MaybeUninit::<[u8; N]>::uninit();
+    // zeroed (not uninit): result_code == 0 is a valid success for variable-size fields
+    // (e.g. an empty SigningPubKey signals a multi-signature transaction), so the buffer
+    // must be valid before the host writes anything.
```
ricky122-5, line 143: Could we potentially refine this? My idea was that this avoids calling assume_init() on something that's uninitialized, not necessarily invalid. Just a doc thing though

ricky122-5 (approved):

xrpl-wasm-stdlib/src/host/field_helpers.rs
sappenin: Not sure I follow your suggestion. Can you propose something specific so I can better understand?

sappenin:
Closing this PR. It was included as part of https://github.com/ripple/xrpl-wasm-stdlib/pull/148

PR #195, opened 2026-06-25
Escrow ctx

ricky122-5:
### Type of Change

- [ ] Documentation update
- [ ] Release


xrpl-escrow-stdlib/src/ctx/finish_result.rs
```diff
@@ -0,0 +1,172 @@
+use core::num::NonZeroI32;
+
+/// Return type for Smart Escrow finish entry points.
+///
+/// Wraps the `i32` the host inspects after the contract returns:
+/// - positive → escrow released
+/// - `0`      → escrow not released, no error (hold)
+/// - negative → finish rejected
+///
+/// Construct with [`FinishResult::succeed`] / [`FinishResult::hold`] /
```
sappenin: I think `FinishResult` should be a generic thing for all Smart Features (not an Escrow-specific thing). 

If desired, we could use a trait like you did with `EscrowFinishContext` and `SmartFeatureContext`? But at a minimum I think we need to design this to be more "Smart Feature" focused, and not (necessarily) Smart Escrow focused.

xrpl-escrow-stdlib/src/ctx/finish_result.rs
```diff
@@ -0,0 +1,172 @@
...
+pub struct FinishResult(i32);
+
+impl FinishResult {
+    /// Release the escrow (maps to `1`).
+    pub const fn succeed() -> Self {
+        Self(1)
+    }
+
+    /// Do not release the escrow; no error (maps to `0`).
+    pub const fn hold() -> Self {
```
sappenin: I don't think `hold` is a verb in either `Smart Features` or `Smart Escrow`. 

For Smart Escrow, it's `finish` or `reject`. For Smart Feature, I think it's probably more broadly just `Succeed` or `Reject`.

sappenin (changes requested):

xrpl-escrow-stdlib/src/ctx/finish_result.rs
ricky122-5: Sounds good. In that case, will bring it out of this PR.

sappenin (approved):

PR #196, opened 2026-06-26
feat: FinishResult for SmartFeatures

ricky122-5:
### Type of Change

- [ ] Documentation update
- [ ] Release


xrpl-wasm-stdlib/src/ctx/finish_result.rs
```diff
@@ -0,0 +1,162 @@
+use core::num::NonZeroI32;
+
+/// Return type for Smart Feature entry points.
+///
+/// Wraps the `i32` the host inspects after a Smart Feature's WASM function
+/// returns:
+/// - positive (`> 0`) → the Smart Feature allows the native transaction to proceed
+/// - `0`              → the Smart Feature blocks the transaction (no specific error)
+/// - negative (`< 0`) → the Smart Feature blocks and propagates a host error code
```
sappenin: I think this error code only has meaning for the contract author as opposed to the host.

```suggestion
/// - negative (`< 0`) → the Smart Feature blocks and propagates a contract-defined error code
```

xrpl-wasm-stdlib/src/ctx/finish_result.rs
```diff
@@ -0,0 +1,162 @@
...
+///
+/// Both `0` and negative values are rejections (per XLS-100: a return value
+/// greater than zero allows the operation, otherwise it is blocked). The
+/// distinction is purely for diagnostics: `0` is a clean rejection, while a
```
sappenin:
```suggestion
/// distinction is purely for diagnostic: `0` is a clean rejection, while a
```

xrpl-wasm-stdlib/src/ctx/finish_result.rs
```diff
@@ -0,0 +1,162 @@
...
+/// negative value carries an error code from a failed host call.
```
sappenin:
```suggestion
/// negative value carries an error code that carries meaning defined by the contract author. For example, returning `-6` maybe me the result  of a failed host call. However, that same value may have a different meaning, defined by the contract author.
```

xrpl-wasm-stdlib/src/ctx/finish_result.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub fn succeed_with(code: NonZeroI32) -> Option<Self> {
+        (code.get() > 0).then_some(Self(code.get()))
+    }
+
+    /// Block the Smart Feature with a custom negative error code.
+    ///
+    /// Returns `None` if `code >= 0`, since a non-negative code would not
+    /// carry an error signal.
+    pub fn reject_with(code: NonZeroI32) -> Option<Self> {
+        (code.get() < 0).then_some(Self(code.get()))
```
sappenin: Is returning an `Option` the right call here? I was discussing this a bit with AI and it suggested clamping any `positive` value to 0. The argument was that the developer is calling `reject_with`, so they clearly don't want to return anything positive, and `0` is a good default.

That said, I could also see a case for a panic here. The contract author supplied a contradictory input.

xrpl-wasm-stdlib/src/ctx/finish_result.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub const fn reject() -> Self {
+        Self(0)
+    }
+
+    /// Allow the Smart Feature with a custom positive code.
+    ///
+    /// Returns `None` if `code <= 0`, since a non-positive code would be
+    /// interpreted by the host as a rejection.
+    pub fn succeed_with(code: NonZeroI32) -> Option<Self> {
+        (code.get() > 0).then_some(Self(code.get()))
```
sappenin: What happens if a contract author returns `succeed_with(-1)`? This function will return an `None` result, but what will the macro or other upstream code do with a `None`?

xrpl-wasm-stdlib/src/ctx/finish_result.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// Block the Smart Feature with a custom negative error code.
+    ///
+    /// Returns `None` if `code >= 0`, since a non-negative code would not
+    /// carry an error signal.
+    pub fn reject_with(code: NonZeroI32) -> Option<Self> {
+        (code.get() < 0).then_some(Self(code.get()))
+    }
+
+    /// Returns `true` if this result allows the transaction to proceed.
+    pub const fn is_succeed(self) -> bool {
```
sappenin: Are these functions useful for anything? The only thing I can think of would be usage by the macro, but not the contract developer. If true, we should consider moving them (or just deleting them).

sappenin (changes requested):

xrpl-wasm-stdlib/src/ctx/finish_result.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// is a compile-time error. The code is recorded in the `WasmReturnCode`
+    /// ledger metadata field and can be used for diagnostics.
+    ///
+    /// ```
+    /// # use xrpl_wasm_stdlib::FinishResult;
+    /// let result = FinishResult::reject_with::<-5>();
+    /// assert_eq!(i32::from(result), -5);
+    /// ```
+    pub fn reject_with<const N: i32>() -> Self {
+        const { assert!(N < 0, "reject_with requires a negative code") };
```
sappenin: A `0` response is still considered a `Reject`:

```suggestion
        const { assert!(N <= 0, "reject_with requires a negative code") };
```

xrpl-escrow-stdlib/src/ctx/finish_result.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn succeed_with_i32_max_uses_code() {
+        assert_eq!(
+            i32::from(FinishResult::succeed_with::<{ i32::MAX }>()),
+            i32::MAX
+        );
+    }
+
+    #[test]
+    fn reject_with_negative_uses_code() {
+        assert_eq!(i32::from(FinishResult::reject_with::<-5>()), -5);
```
sappenin, line 113: Let's test each `succeed_with` and `reject_with` using `0`.

sappenin (changes requested):

xrpl-wasm-stdlib/src/ctx/finish_result.rs
ricky122-5: I agree with this, I originally thought that the 0 case was reserved for the basic reject() method, but giving the freedom to reject with 0 is a good catch, thanks!

xrpl-escrow-stdlib/src/ctx/finish_result.rs
ricky122-5: succeed_with(0) is a compile time error - are you suggesting a trybuild test?

sappenin: Oh, good point - yeah, since we're using `trybuild` anyway, this seems like a good test case. WDYT?

xrpl-escrow-stdlib/src/ctx/finish_result.rs
```diff
@@ -0,0 +1,155 @@
...
+///
+/// Construct with [`FinishResult::succeed`] / [`FinishResult::reject`] for the
+/// common cases, or [`FinishResult::succeed_with`] / [`FinishResult::reject_with`]
+/// for a custom code. `From<i32>` is also implemented so error codes from host
+/// calls (e.g. `e.code()`) can be propagated directly:
+/// `return e.code().into();`
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct FinishResult(i32);
+
+impl FinishResult {
```
sappenin, line 24: I'm not sure this struct is in the right module. Let's discuss that before we merge this one.

xrpl-escrow-stdlib/src/ctx/finish_result.rs
ricky122-5: I like it, will add

xrpl-escrow-stdlib/tests/finish_result/compile_fail/succeed_with_negative.rs
```diff
@@ -0,0 +1,5 @@
+use xrpl_escrow_stdlib::FinishResult;
```
sappenin, line 1: The folder named `compile_fail` isn't descriptive. What if we call this `test/finish_result/compile_fail`?

xrpl-escrow-stdlib/src/ctx/finish_result.rs
sappenin: LGTM - see my other comment about folder naming.

xrpl-escrow-stdlib/tests/finish_result/compile_fail/succeed_with_negative.rs
ricky122-5: Fixed, thanks!

sappenin (approved):

PR #198, opened 2026-06-29
feat: Contract crate

ricky122-5:
### Type of Change

- [ ] Documentation update
- [ ] Release

Implement contract crate, etc.


xrpl-wasm-stdlib/src/core/types/amount.rs
```diff
@@ -127,15 +130,15 @@ impl Amount {
...
-                let abs_drops = num_drops.unsigned_abs();
-                let mut value = abs_drops;
                 if *num_drops >= 0 {
-                    value |= 0x4000000000000000u64; // Set bit 6 for positive
+                    // Set the positive bit (0x40) and clear the currency bit (0x80)
+                    let positive_drops = (*num_drops as u64) | 0x4000000000000000u64; // Set positive bit
+                    bytes[0..8].copy_from_slice(&positive_drops.to_be_bytes());
+                } else {
+                    // Clear the positive bit
+                    let negative_drops = ((-*num_drops) as u64) & 0x3FFFFFFFFFFFFFFFu64; // Clear positive bit
```
sappenin, line 136: I don't think this is correct. Even for negative amounts, STAmount's XRP amount bytes are always a positive number; the leading flag is the thing that indicates negative vs. positive.

xrpl-wasm-stdlib/src/core/types/amount.rs
```diff
@@ -190,7 +193,8 @@ impl Amount {
     pub fn from_bytes(bytes: &[u8]) -> host::Result<Self> {
         // TODO: Move to trait!
 
-        if bytes.len() != 48 {
+        // XRP = 8 bytes, MPT = 33 bytes, IOU = 48 bytes
+        if bytes.len() != 8 && bytes.len() != 33 && bytes.len() != 48 {
```
sappenin: This is not correct. STAmount payloads are always 48 bytes. For XRP or MPT (which don't require 48 bytes), the buffer is still 48 bytes but trailing bytes are not read for purposes of parsing.

xrpl-wasm-stdlib/src/core/types/amount.rs
```diff
@@ -247,16 +253,15 @@ impl Amount {
         }
         // is_iou
         else {
-            // If we get here, we'll have 48 bytes.
+            // IOU requires exactly 48 bytes
+            if bytes.len() != 48 {
```
sappenin: This is not needed. See the check on line 197 (or previously, on line 193)

xrpl-wasm-stdlib/src/core/types/amount.rs
```diff
@@ -276,6 +281,120 @@ impl Amount {
             Ok(amount)
         }
     }
+
+    pub fn transfer(&self, recipient: &AccountID) -> i32 {
```
sappenin: Note: This `transfer` function seems out of place on the `amount` struct.

sappenin (changes requested):

xrpl-wasm-stdlib/src/core/types/amount.rs
ricky122-5: I think the comment is a bit misleading but doesn't this clear the IOU sign bit and store the absolute magnitude ( because of -*num_drops)? Not sure if this is incorrect. This is also unmodified code from /contract.

xrpl-wasm-stdlib/src/core/types/amount.rs
ricky122-5: Moved to contract/

xrpl-wasm-stdlib/src/core/types/amount.rs
ricky122-5: Fixed, thanks!

xrpl-wasm-stdlib/src/core/types/amount.rs
ricky122-5: Fixed, thanks!

mvadari:
@dangell7 would you mind taking a quick look at this?

e2e-tests/smart-contracts/amount_tests/src/lib.rs
```diff
@@ -0,0 +1,244 @@
+#![cfg_attr(target_arch = "wasm32", no_std)]
```
mvadari, line 1: Most of these e2e tests look contract-specific - they should go into a separate subfolder called `smart-contracts` or something.

xrpl-contract-stdlib/xrpl-parameter-macro/src/lib.rs
```diff
@@ -0,0 +1,170 @@
+extern crate std;
```
mvadari, line 1: Should this be part of the existing macro collection?

xrpl-wasm-stdlib/src/host/host_bindings_empty.rs
```diff
@@ -164,4 +164,20 @@ export_host_functions! {
     fn trace_opaque_float(_msg_read_ptr: *const u8, _msg_read_len: usize, _opaque_float_ptr: *const u8, _opaque_float_len: usize) -> i32;
     fn trace_amount(_msg_read_ptr: *const u8, _msg_read_len: usize, _amount_ptr: *const u8, _amount_len: usize) -> i32;
 
+    // Other functions
+    fn instance_param(_index: i32, _st_type_id: i32, _out_buff_ptr: *mut u8, _out_buff_len: usize) -> i32;
```
mvadari, line 168: This seems to be against one of our core goals of "you can't call a smart contract host function from a smart escrow"

xrpl-wasm-stdlib/src/sflags.rs
```diff
@@ -0,0 +1,228 @@
+// Find a way to auto generate these from the ripple(d) source code
+
+#![allow(non_upper_case_globals)]
```
mvadari, line 4: There should be a script to keep this updated

xrpl-wasm-stdlib/src/core/types/number.rs
```diff
@@ -0,0 +1,30 @@
+// use crate::core::types::amount::opaque_float::OpaqueFloat;
```
mvadari, line 1: Adding these new types to the common lib should be a separate PR, so that we can properly test everything etc

xrpl-wasm-stdlib/src/core/type_codes.rs
```diff
@@ -0,0 +1,32 @@
+// Find a way to auto generate these from the ripple(d) source code
+
+// pub const STI_UNKNOWN: u8 = -2;
+// pub const STI_NOTPRESENT: u8 = 0;
+pub const STI_UINT16: u8 = 1;
```
mvadari, line 7: There should be a script to generate this list

xrpl-wasm-stdlib/src/core/types/amount.rs
sappenin: Oh, yes - my mistake. The variable `negative_drops` confused me.

Is this function broken though? I was under the impression it's working correctly currently. If it's broken, we should fix it independently from the `contract` PR (I think).

xrpl-wasm-stdlib/src/host/host_bindings_empty.rs
ricky122-5: It's only accessible via unsafe {} which is the same case vice versa. In the design doc the host function split is done by the EscrowFinishContext / ContractCallContext objects having a type-safe wrapper over the host function. Ideally by the end we shouldn't have any cases of unsafe calls, and if so, rippled will reject it

Should ContractCallContext be added in this PR? I'm open to adding it but this PR was originally geared towards merging the existing contract branch in

xrpl-contract-stdlib/xrpl-parameter-macro/src/lib.rs
ricky122-5: I think eventually. Right now we have an empty smart_contract macro stub. The behavior of this macro is similar can be absorbed into it, but I'd think that would come when the smart_contract tickets are being done since this macro will be eventually removed. I think it would be better to remove when the time comes, but open to other ideas.

xrpl-wasm-stdlib/src/core/type_codes.rs
ricky122-5: Added in latest commit, thanks

xrpl-wasm-stdlib/src/sflags.rs
ricky122-5: Added in latest commit, thanks

mvadari (changes requested):
This PR is way too big to be reviewed. PRs should ideally be under 1k lines, unless they're super mechanical. Start by splitting out the following each into their own separate PR:
* Adding the flags and the script to populate them
* The type changes/additions in the common library

ricky122-5:
Closed, split into 8 PRs starting with https://github.com/ripple/xrpl-wasm-stdlib/pull/224

PR #199, opened 2026-06-29
refactor: Make examples use escrow crate 

ricky122-5:
### Type of Change

- [ ] Documentation update
- [ ] Release


examples/smart-escrows/atomic_swap/atomic_swap1/src/lib.rs
```diff
@@ -3,20 +3,20 @@
...
-use xrpl_wasm_stdlib::core::types::contract_data::XRPL_CONTRACT_DATA_SIZE;
-use xrpl_wasm_stdlib::host;
-use xrpl_wasm_stdlib::host::Error::InternalError;
-use xrpl_wasm_stdlib::host::error_codes::match_result_code_with_expected_bytes;
-use xrpl_wasm_stdlib::host::get_tx_nested_field;
-use xrpl_wasm_stdlib::host::trace::{DataRepr, trace_data, trace_num};
-use xrpl_wasm_stdlib::host::{Error, Result, Result::Err, Result::Ok};
-use xrpl_wasm_stdlib::sfield;
-use xrpl_wasm_stdlib::types::{ContractData, XRPL_CONTRACT_DATA_SIZE as TX_CONTRACT_DATA_SIZE};
+use xrpl_escrow_stdlib::core::keylets::XRPL_KEYLET_SIZE;
```
sappenin: These are a bit surprising to me. Many of these aren't `Smart Escrow` specific, but are instead slightly more generic (i.e., will apply to all Smart Features at a minimum, and likely will will apply to SC). It seems like the correct place for these is actually just `xrpl_wasm_stdlib`.

As an example, imagine we have a Smart MPT in the future. I don't think we should have another import path for common primitives like this:

```
use xrpl_smart_mpt_stdlib::host::get_tx_nested_field;
```

If we do that, we'll be duplicating common things across every smart feature.

mvadari (changes requested):
This PR is incorrectly named - if it's a fix, it's not a feat (this actually looks like it should be a refactor) - and also, what are you fixing?

examples/smart-escrows/atomic_swap/atomic_swap1/src/lib.rs
ricky122-5: Those live in the common stdlib, but the escrow stdlib just reimports them. The end goal is to have a single global import per feature.

```rust
use xrpl_smart_mpt_stdlib::*
```
or 
```rust
use xrpl_smart_escrow_stdlib::*
```

and get both escrow-specific and common types

sappenin: I see. I don't love that pattern (if I'm understanding it correctly).

For primitives that are _not_ escrow-specific, it's important to maintain the current import scheme so that we don't accidentally imply to a contract developer that core primitives vary by Smart Feature. This will also help smart feature developers internalize a single concept for core primitives. 

Based on your comment above, it seems like one motivation for this  proposed change is that you were also hoping to flatten the use/import structure? I.e., flatten things like `core::locator`, `core::types`, `host::trace`, etc into a single wildcard? 

This is generally discouraged because while wildcards are convenient, heavy use of `*` (outside of a prelude module or tests) is generally discouraged due to namespace pollution (i.e., it makes it easy to accidentally introduce name clashes). We already we have name clashes in the SDK (e.g.,  between host functions and helper functions with the same name), so I don't think a wildcard import will work even if we wanted it to. 

Another thing that suffers is just readability (i.e., when someone reads your code later, it’s much harder to tell where a function like trace() actually came from)

Does my suggestion to not collapse the namespace break things for your overall design? If so, let's talk!

PR #202, opened 2026-06-30
refactor: Convert some internal error usages to panic

pwang200:
Note this PR is before the Observability improvements refactoring. So it is assuming -1 is still the internal error. 

Convert internal-invariant trips to panics; reclassify input errors                        
                                                                                             
  Some sites returned Err(InternalError) for conditions that are either (a) genuine          
  invariant violations the caller can't recover from, or (b) plain input errors              
  mislabeled as internal. This cleans them up:                                               
                                                                                             
  Panic (true invariant trips — unreachable unless the host/stdlib is buggy):                
  - error_codes.rs `match_result_code_with_expected_bytes`: host returned a                  
    non-negative byte count that doesn't match the expected size.                            
  - current_tx/traits.rs `get_signing_pub_key`: SigningPubKey length other than              
    0 or 33. rippled preflight rejects such txns (temBAD_SIGNATURE) before they              
    execute, so a running escrow can never observe this.                                     
                                                                                             
  Reclassify (caller/input errors, not internal):                                            
  - amount.rs `from_bytes`: wrong-length slice → InvalidParams.                              
  - example escrows nft_owner/atomic_swap1: empty memo (host returns 0) is                   
    protocol-valid input → Ok(None), matching the Option return type and existing            
    caller handling.                                                                         
  - example escrow freelancer: empty memo → InvalidParams (its signature has no              
    Option; consistent with how it already rejects an unrecognized intent byte).             
                                                                                             
  Tests updated to #[should_panic] / assert the new codes.                                   
                                                                                             
  Why the remaining InternalError usages are unchanged:                                      
  They are not stdlib decisions — they are (1) the const/enum definition, which must         
  stay because the host itself can still return -1, propagated via Error::from_code,         
  and (2) tests that simulate the host returning INTERNAL_ERROR and assert the stdlib        
  propagates it (correct: host-reported internal errors should surface as Err, not           
  panic). No production site constructs InternalError directly anymore.

### DIFF
--- a/xrpl-wasm-stdlib/src/core/current_tx/traits.rs
+++ b/xrpl-wasm-stdlib/src/core/current_tx/traits.rs
@@ -233,7 +233,13 @@ pub trait TransactionCommonFields {
     /// Returns a `Result<Option<PublicKey>>` where:
     /// * `Ok(Some(PublicKey))` - The 33-byte compressed public key for single-signature transactions
     /// * `Ok(None)` - Empty SigningPubKey field, indicating a multi-signature transaction
-    /// * `Err(Error)` - If the field cannot be retrieved or has an unexpected size (not 0 or 33 bytes)
+    /// * `Err(Error)` - If the field cannot be retrieved
+    ///
+    /// # Panics
+    ///
+    /// Panics if the field is present with a length other than 0 or 33 bytes. rippled's
+    /// preflight rejects such transactions before they are applied, so this is an internal
+    /// invariant violation rather than recoverable input.
     ///
     /// # Security Note
     ///
@@ -244,9 +250,8 @@ pub trait TransactionCommonFields {
         get_field(sfield::SigningPubKey).and_then(|blob| match blob.len {
             0 => Result::Ok(None), // Multi-signature transaction
             33 => Result::Ok(Some(PublicKey::from(blob.data))), // Single-signature transaction
-            _ => Result::Err(crate::host::Error::from_code(
-                crate::host::error_codes::INTERNAL_ERROR,
-            )),
+            // Unreachable in practice (see `# Panics`); fail fast if the invariant breaks.
+            len => panic!("internal invariant violated: SigningPubKey has unexpected length {len} (expected 0 or 33)"),
         })
     }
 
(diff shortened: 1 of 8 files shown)


xrpl-wasm-stdlib/src/host/error_codes.rs
```diff
@@ -149,7 +154,12 @@
 {
     match result_code {
         code if code as usize == expected_num_bytes => Ok(on_success()),
-        code if code >= 0 => Err(InternalError), // If here, this is a bug
+        // Non-negative but wrong byte count: internal invariant violation (see `# Panics`).
+        code if code >= 0 => {
+            panic!(
```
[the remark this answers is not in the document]
pwang200: keylet is public information.

xrpl-wasm-stdlib/src/core/current_tx/traits.rs
```diff
@@ -244,9 +250,8 @@ pub trait TransactionCommonFields {
         get_field(sfield::SigningPubKey).and_then(|blob| match blob.len {
             0 => Result::Ok(None), // Multi-signature transaction
             33 => Result::Ok(Some(PublicKey::from(blob.data))), // Single-signature transaction
-            _ => Result::Err(crate::host::Error::from_code(
-                crate::host::error_codes::INTERNAL_ERROR,
-            )),
+            // Unreachable in practice (see `# Panics`); fail fast if the invariant breaks.
+            len => panic!("internal invariant violated: SigningPubKey has unexpected length {len} (expected 0 or 33)"),
```
sappenin, line 254: I don't think any text from a `panic!` ever gets seen (maybe double-check with a local rippled to see?). My current impression is you need to `trace` the error message before calling panic if you want this to show up in logs.

Can you confirm either way?

xrpl-wasm-stdlib/src/core/current_tx/traits.rs
```diff
@@ -1109,30 +1147,38 @@ mod tests {
 
                 let tx = EscrowFinish;
 
-                // Fixed-size mandatory fields should return Err when zero length (byte mismatch)
-                let account_result = tx.get_account();
-                assert!(account_result.is_err());
```
sappenin, line 1114: I think we should restore these tests. There are actually two things going on in this test. First is, "do variable size fields return OK when 0-length, in the general case (for that, we can just check a single filed, like `tx.get_fee()`, as you do)

However, a second thing being covered here is "do specific functions do this correctly" (e.g., tx.get_transaction_type()` and the rest that were removed).

xrpl-wasm-stdlib/src/host/error_codes.rs
```diff
@@ -149,7 +154,12 @@ where
...
```
sappenin, line 159: See my comment above around tracing the error messages.

sappenin (changes requested):

xrpl-wasm-stdlib/src/core/current_tx/traits.rs
pwang200: With wasm as compilation target, the panic message is dropped. 

Claude suggested the `trace` calls. But I deleted them. It is good to have more detailed debug information, and the gas cost is gonna be low and should never be called. On the other hand, adding those `trace` calls will result in extra bytecode that all contracts need to carry. I picked smaller code size.

xrpl-wasm-stdlib/src/core/current_tx/traits.rs
pwang200: I think the github diff page is confusing. If the PR is opened to see the diff side by side, it'd be easy to see all the test cases are still there, but split to different tests because the first panic case will terminate the test, so different panic cases must be tested one by one.

xrpl-wasm-stdlib/src/core/current_tx/traits.rs
sappenin: Ah, makes sense. I was just thinking about the developer experience, but I think a dev can see a line number after a panic, right (I mean, during development. If this happens on a live network, even the trace won't help)

xrpl-wasm-stdlib/src/core/current_tx/traits.rs
sappenin: Oh, I see where they go moved to. Thanks for pointing that out.

sappenin (approved):

PR #204, opened 2026-06-30
refactor: Move escrow code to crate, change imports

ricky122-5:

Sketch of the escrow file move if we were to go with the proposal of moving escrow-specific code to the escrow crate instead of importing from common.

### Type of Change

- [ ] Documentation update
- [ ] Release


xrpl-escrow-stdlib/src/current_tx/escrow_finish.rs
```diff
@@ -0,0 +1,16 @@
+use xrpl_wasm_stdlib::core::current_tx::traits::TransactionCommonFields;
+
+use crate::current_tx::traits::EscrowFinishFields;
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+#[repr(C)]
```
sappenin: I don't think we need `#[repr(C)]` here since an `EscrowFinish` never cross the FFI boundary (I think that's correct based on your other PRs). By default, Rust uses `#[repr(Rust)]`, which grants the compiler full permission to optimize the memory layout of the struct. Forcing `#[repr(C)]` strips away these optimizations.

We should probably audit usage of this macro in other places just to make sure we want to keep it.

xrpl-escrow-stdlib/src/ledger_objects/escrow.rs
```diff
@@ -19,14 +20,3 @@ impl Escrow {
         Self { slot_num }
     }
 }
-
-#[cfg(test)]
-mod tests {
```
sappenin, line 24: Any reason to remove this test?

sappenin (changes requested):

docs/comprehensive-guide.md
```diff
@@ -100,9 +100,9 @@ npm install
 
 Let's create a simple escrow that releases funds when an account balance exceeds 10 XRP:
 
-```rust
+```rust ignore
```
mvadari, line 103: Why are these marked as `ignore` now?

e2e-tests/gas_benchmark/Cargo.toml
```diff
@@ -10,3 +10,4 @@ crate-type = ["cdylib"]
 
 [dependencies]
 xrpl-wasm-stdlib = { path = "../../xrpl-wasm-stdlib" }
+xrpl-escrow-stdlib = { path = "../../xrpl-escrow-stdlib" }
```
mvadari, line 13: Do you need both dependencies or is one enough?

examples/smart-escrows/atomic_swap/atomic_swap1/src/lib.rs
```diff
@@ -17,6 +14,10 @@ use xrpl_wasm_stdlib::host::{Error, Result, Result::Err, Result::Ok};
 use xrpl_wasm_stdlib::sfield;
 use xrpl_wasm_stdlib::types::{ContractData, XRPL_CONTRACT_DATA_SIZE as TX_CONTRACT_DATA_SIZE};
 
+use xrpl_escrow_stdlib::ledger_objects::current_escrow::{self, CurrentEscrow};
```
mvadari: This is confusing linting-wise - why is it only below `xrpl_wasm_stdlib` here?

docs/comprehensive-guide.md
ricky122-5: These wouldn't compile with the escrow stdlib as docs/comprehensive-guide uses a symlink with common's lib.rs. One possible solution is a dev-dependency but that would probably need more discussion

e2e-tests/gas_benchmark/Cargo.toml
ricky122-5: Yep, both are needed bc it uses Escrow fields as well as TransactionCommonFields (examples) that are from both common and escrow stdlibs.

examples/smart-escrows/atomic_swap/atomic_swap1/src/lib.rs
ricky122-5: I grouped them this way for readability. Open to just merge into one block if you think it would be better. No strong opinion on this

docs/comprehensive-guide.md
mvadari: Can it not use a symlink for the escrow stdlib?

e2e-tests/gas_benchmark/Cargo.toml
mvadari: Right, but aren't the common fields included in the escrow-stdlib?

examples/smart-escrows/atomic_swap/atomic_swap1/src/lib.rs
mvadari: Be consistent - pick one setup and stick with it. I'd recommend still doing alphabetical order, so keeping `xrpl_escrow_stdlib` above `xrpl_wasm_stdlib`.

e2e-tests/gas_benchmark/Cargo.toml
ricky122-5: Actually I left in the wildcard import on accident, will fix in the upcoming commit

However escrow-stdlib and common-stdlib aren't supposed to have overlap. I believe we discussed when the wildcard import came up, and I think we came to the conclusion where all escrow code lives in its stdlib, and common code lives in its stdlib.

docs/comprehensive-guide.md
ricky122-5: I could, but the comprehensive guide module would also have to move to lib.rs in the escrow crate if that's okay. I think this could be fine because I don't think the guide has to really do anything with things other than escrow, but it's something to keep in mind if/when we add new smart features (adding those examples may fail)

Or we could give it its own crate maybe?

xrpl-escrow-stdlib/src/ledger_objects/escrow.rs
ricky122-5: fixed

examples/smart-escrows/atomic_swap/atomic_swap1/src/lib.rs
ricky122-5: Gotcha, gonna have one block in alphabetical order for all examples

sappenin (approved):

docs/comprehensive-guide.md
mvadari: Why does a symlink not work for that?

xrpl-wasm-stdlib/src/core/ledger_objects/account_root.rs
```diff
@@ -6,7 +6,6 @@ use crate::host;
 use host::Error;
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
-#[repr(C)]
```
mvadari, line 9: This seems like an unrelated change

ricky122-5: David's comment addressed it earlier, fixed based on that bc AccountRoot just is a slot number wrapper without crossing the FFI. I included it for consistency with the escrow_finish changes because AccountRoot is similar

xrpl-wasm-stdlib/src/core/ledger_objects/traits.rs
```diff
@@ -915,7 +617,7 @@ mod tests {
 
             let _guard = setup_mock(mock);
 
-            let escrow = CurrentEscrow;
+            let escrow = TestCurrentLedgerObject;
```
mvadari: variable names should be fixed here

xrpl-escrow-stdlib/src/ledger_objects/traits.rs
```diff
@@ -0,0 +1,742 @@
...
+    Error, get_current_ledger_obj_field, get_ledger_obj_field, update_data,
+};
+use xrpl_wasm_stdlib::host::{Result, Result::Err, Result::Ok};
+use xrpl_wasm_stdlib::sfield;
+
+/// Trait providing access to fields specific to Escrow objects in the current ledger.
+///
+/// This trait extends `CurrentLedgerObjectCommonFields` and provides methods to access
+/// fields that are specific to Escrow objects in the current ledger being processed.
+pub trait CurrentEscrowFields: CurrentLedgerObjectCommonFields {
```
mvadari, line 19: This shouldn't be moved from the common stdlib, you can still access escrow from e.g. smart contracts

sappenin: I agree with the comment, but only for `EscrowFields` and `LedgerObjectCommonFields`.

For `CurrentEscrowFields` and `CurrentLedgerObjectCommonFields`, I think I disagree. Assuming Smart Contracts will use cache_slots and locators in the same way Smart Escrows do, the traits that Smart Escrows will want to use are instead `EscrowFields` and `LedgerObjectCommonFields`.

`CurrentEscrowFields` and `CurrentLedgerObjectCommonFields` define functions that don't use a slot (because the location is assumed by context). All that to say, I don't think a Smart Contract will be able to use those, so those should move here.

xrpl-wasm-stdlib/src/core/ledger_objects/traits.rs
```diff
@@ -92,299 +85,6 @@ pub trait CurrentLedgerObjectCommonFields {
...
-        match_result_code(result_code, || ())
-    }
-}
-
-/// Trait providing access to fields specific to Escrow objects in any ledger.
-///
-/// This trait extends `LedgerObjectCommonFields` and provides methods to access
-/// fields that are specific to Escrow objects in any ledger, not just the current one.
-/// Each method requires a register number to identify which ledger object to access.
-pub trait EscrowFields: LedgerObjectCommonFields {
```
sappenin, line 240: Per the above, I think this is a mistake and should stay here.

ricky122-5: Thanks, updated

xrpl-wasm-stdlib/src/core/ledger_objects/traits.rs
ricky122-5: Renamed to `obj	`, thanks

sappenin (approved):

mvadari (approved):

mvadari (approved):

PR #206, opened 2026-06-30
Update Amount to work with XFloat

sappenin:
- Amount IOU field simplification: Amount::IOU.amount changes from OpaqueFloat (a newtype wrapper) to [u8; 8], removing the unnecessary wrapper. Adds Amount::to_xfloat(rounding_mode) and Amount::iou_from_mant_exp(...) helpers for converting between Amount and XFloat via the host.
  - Typed RoundingMode enum: replaces the four raw FLOAT_ROUNDING_MODES_* i32 constants with a #[repr(i32)] enum whose variants map exactly to rippled's Number::rounding_mode C++ enum values (verified by unit tests). All float host calls (XFloat::from_mant_exp, Amount::to_xfloat — now require an explicit RoundingMode argument.)

### DIFF
--- a/xrpl-wasm-stdlib/src/core/types/xfloat.rs
+++ b/xrpl-wasm-stdlib/src/core/types/xfloat.rs
@@ -1,130 +1,56 @@
-/// Opaque 96-bit representation of an XRPL fungible token (IOU) amount (STNumber format).
+use crate::host;
+use crate::host::Result;
+use crate::host::RoundingMode;
+
+/// Opaque 96-bit (12-byte) float in rippled's STNumber wire format.
 ///
-/// This struct encapsulates the 12-byte STNumber format used by rippled's host functions.
-/// Layout: 8-byte big-endian int64 mantissa followed by 4-byte big-endian int32 exponent.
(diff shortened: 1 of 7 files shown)


xrpl-wasm-stdlib/src/core/types/xfloat.rs
```diff
@@ -1,130 +1,56 @@
...
+    ///
+    /// Creates an `XFloat` from a mantissa and exponent via the host's `float_from_mant_exp`.
+    ///
+    /// `rounding_mode` controls how the host rounds the result.
+    pub fn from_mant_exp(
+        mantissa: i64,
+        exponent: i32,
+        rounding_mode: RoundingMode,
+    ) -> Result<Self> {
+        let mut float_out = [0u8; 12];
```
sappenin, line 42: NOTE: This is incorrect at present. While the `amount` on an IOU is 8 bytes, we can't take an `XFloat` (`STNumber`) and convert it to an `IOUAmount` without a host function interaction. 

Gist proposal coming soon.

PR #207, opened 2026-07-02
feat: Smart_escrow entry point macro

ricky122-5:

Adds the #[smart_escrow] macro which validates and wraps a valid user function with the required extern "C" fn finish() -> i32 wrapper. This allows the user freedom in function names and structure.
Also edits the hello_world example to use this macro.

### Type of Change

- [ ] Documentation update
- [ ] Release


kuznetsss:
Please, update the examples in the doc: https://github.com/ripple/xrpl-wasm-stdlib/blob/main/docs/comprehensive-guide.md
It looks like at least this [section](https://github.com/ripple/xrpl-wasm-stdlib/blob/main/docs/comprehensive-guide.md#your-first-contract) could use the new macro.

xrpl-wasm-stdlib/src/lib.rs
```diff
@@ -10,6 +10,8 @@ pub use xrpl_macros::currency;
 pub use xrpl_macros::hash256;
 pub use xrpl_macros::pubkey;
 pub use xrpl_macros::r_address;
+pub use xrpl_macros::smart_contract;
+pub use xrpl_macros::smart_escrow;
```
kuznetsss, line 14: Shouldn't `smart_escrow` macro be a part of `xrpl-escrow-stdlib`? I don't think it is needed in the general purpose stdlib

xrpl-wasm-stdlib/src/lib.rs
```diff
@@ -10,6 +10,8 @@ pub use xrpl_macros::currency;
 pub use xrpl_macros::hash256;
 pub use xrpl_macros::pubkey;
 pub use xrpl_macros::r_address;
+pub use xrpl_macros::smart_contract;
```
kuznetsss, line 13: `smart_contract` macro doesn't do anything now. Maybe `smart_contract` should create an entry point for a smart contract while `smart_escrow` should check signature and convert `FinishResult` into `i32` for smart escrow smart contracts only?

xrpl-wasm-stdlib/src/lib.rs
ricky122-5: Later on we're breaking away the xrpl_macros crate from xrpl_wasm_stdlib as its own crate at the root level. Regardless, smart_escrow will need to be in a different crate because it's a proc_macro crate. We could put it in xrpl-escrow-stdlib but it would do a similar thing (get re-exported by the crate) and needs discussion if it's worth adding a new crate for one macro

xrpl-wasm-stdlib/src/lib.rs
ricky122-5: Both smart_contract and smart_escrow macros create a wrapper on functions to omit the boilerplate. smart_contract just hasn't been implemented yet as it's stdlib implementation is still in the contract branch

ricky122-5:
> Please, update the examples in the doc: https://github.com/ripple/xrpl-wasm-stdlib/blob/main/docs/comprehensive-guide.md It looks like at least this [section](https://github.com/ripple/xrpl-wasm-stdlib/blob/main/docs/comprehensive-guide.md#your-first-contract) could use the new macro.

Thanks! Added in https://github.com/ripple/xrpl-wasm-stdlib/pull/208

xrpl-wasm-stdlib/src/lib.rs
kuznetsss: I see. If `xrpl_macros` will be moved out from `xrpl_wasm_stdlib` then I think it is fine to keep `smart_escrow` macro in it

kuznetsss:
> Thanks! Added in https://github.com/ripple/xrpl-wasm-stdlib/pull/208

Thank you!
But maybe you could move docs update into this PR? So all the changes related to adding `smart_escrow` macro would be in one PR

docs/comprehensive-guide.md
```diff
@@ -170,8 +175,11 @@ Smart escrows are **conditional payment contracts** that:
 
 Every smart escrow must:
 
-1. **Export a `finish()` function** with signature `extern "C" fn finish() -> i32`
-2. **Return 1 to release** funds or **0 to keep locked**
+1. **Export a `finish()` function** with signature `extern "C" fn finish() -> i32`. The `#[smart_escrow]` macro
+   generates this for you from a function you annotate — see [Your First Contract](#your-first-contract) — so you
+   normally never write the raw `extern "C"` export yourself.
```
kuznetsss, line 180: How do you think, maybe we don't need, to explain implementation details for users?
I mean, maybe we could simply say: annotate smart contract function with `#[smart_escrow]` macro.

docs/comprehensive-guide.md
```diff
@@ -170,8 +175,11 @@ Smart escrows are **conditional payment contracts** that:
...
+2. **Return a positive value to release** funds, or **zero/negative to keep locked** (`FinishResult::succeed()` /
+   `FinishResult::reject()` under the macro)
```
kuznetsss, line 182: Here maybe we also don't need to provide implementation details?
```suggestion
2. **Return a `FinishResult::succeed()` value to release** funds, or **`FinishResult::reject()` to keep locked** 
```

docs/comprehensive-guide.md
sappenin: We should document, in at least one place in this doc, the mapping from XLS-101's `finish` function ([[here](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0100-smart-escrows#141-notary-release)](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0100-smart-escrows#141-notary-release)) to the stdlib's Rust structs. I don't see this described elsewhere, so this spot seems fine.

Open to shortening this (or just linking XLS-101) as long as the mapping is described somewhere.

docs/comprehensive-guide.md
sappenin: I think we should keep this as-is. It's probably still important to speak in terms of positive and negative values. That's the core primitive, and `FinishResult` allows more than just `succeed()` and `reject()` (e.g., see `reject_with()`).

sappenin (approved):

sappenin (approved):

sappenin (approved):

PR #211, opened 2026-07-07
feat: FieldDecoder, marker traits

ricky122-5:

Adds the FieldDecoder trait, marker traits, and field_source macro as setup for replacing the LedgerObjGetter and CurrentTxGetter traits.

### Type of Change

- [ ] Documentation update
- [ ] Release

Added unit tests and trybuild tests (for incorrect context compile fails, e.g trying to get a CurrentTx-only field in a LedgerObj context)


xrpl-wasm-stdlib/Cargo.toml
```diff
@@ -29,3 +29,4 @@ mockall = { version = "0.14.0", optional = true }
 
 [dev-dependencies]
 mockall = "0.14.0"
+trybuild = "1"
```
mvadari: This should have the full version number

xrpl-wasm-stdlib/src/fields/decoder.rs
```diff
@@ -0,0 +1,166 @@
...
+///
+/// # Examples
+///
+/// ```ignore
+/// field_source!(MyType : tx);        // decodable from the current transaction only
+/// field_source!(MyType : obj);       // decodable from a ledger object only
+/// field_source!(MyType : tx, obj);   // decodable from either source
+/// ```
+#[allow(unused_macros)]
+macro_rules! field_source {
```
mvadari: I would suggest integrating FieldDecoder in at least one place in this PR, so you know that it all hooks together correctly, and save the macro for a separate PR - given that this is a helper macro, we may decide that we'd rather have the non macro version.

ricky122-5: It's probably not possible in this PR unless I expand scope significantly to implementing getters for current_tx and ledger_obj. I feel the tests have enough proof that it should work. I agree that the macro can come in later though. Curious about your thoughts if the getters should be included in this PR

sappenin (approved):

PR #213, opened 2026-07-07
refactor: Common reorg

ricky122-5:

Refactors the common stdlib into distinguishable submodules.

### Type of Change

- [ ] Documentation update
- [ ] Release


xrpl-common-stdlib/src/fields/current_tx/mod.rs
```diff
@@ -76,9 +76,9 @@ use crate::sfield::SField;
 /// ## Usage Patterns
 ///
 /// ```rust,no_run
-/// use xrpl_common_stdlib::core::current_tx::{get_field, get_field_optional};
-/// use xrpl_common_stdlib::core::types::account_id::AccountID;
-/// use xrpl_common_stdlib::core::types::amount::Amount;
+/// use xrpl_common_stdlib::fields::current_tx::{get_field, get_field_optional};
```
sappenin: The `current_tx` isn't really a "field" is it?

xrpl-common-stdlib/src/fields/current_tx/traits.rs
sappenin, line 58: This feels misplaced. Shouldn't this stay in `core` or maybe `core::transactions`?

The `Current TX` is its own concept that's quite a bit narrower than "all txs."

xrpl-common-stdlib/src/host/trace.rs
```diff
@@ -210,9 +210,9 @@ mod tests {
         let _guard = setup_mock(mock);
 
         // Create a test IOU Amount
-        use crate::core::types::account_id::AccountID;
```
sappenin, line 213: This is correct currently, but has me wondering if we wouldn't benefit from `host` and `guest` types. E.g., host types might model just bytes, whereas `guest` types might be more "rusty." Potentially the `EscrowFinishResult` comes to mind as an example. But I wonder if potentially something like an `Amount` might be in this category too. Like, potentailly `STAmount` and `Amount` are two different Rust types, but the WASM contract developer maybe mostly just deals with the `guest` types.

@kuznetsss this might inform our conversation on Monday.

xrpl-common-stdlib/src/types/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+pub mod object;
+pub mod opaque_float;
+pub mod public_key;
+pub mod transaction_type;
+pub mod uint;
+
+// TODO: Move these to the `types` crate.
+// Relocated from the old top-level `src/types.rs`, which collided with this
+// directory's module path once `core::types` was promoted to `crate::types`.
+pub const XRPL_CONTRACT_DATA_SIZE: usize = 4096; //TODO size??
```
sappenin, line 20: Let's verify the contract size. Also, what's the plan for these two lines (20 and 21)?

sappenin (changes requested):

xrpl-common-stdlib/src/current_tx/mod.rs
```diff
@@ -76,9 +76,9 @@ use crate::sfield::SField;
 /// ## Usage Patterns
 ///
 /// ```rust,no_run
-/// use xrpl_common_stdlib::core::current_tx::{get_field, get_field_optional};
-/// use xrpl_common_stdlib::core::types::account_id::AccountID;
-/// use xrpl_common_stdlib::core::types::amount::Amount;
+/// use xrpl_common_stdlib::fields::current_tx::{get_field, get_field_optional};
+/// use xrpl_common_stdlib::types::account_id::AccountID;
```
sappenin, line 80: After discussing with @ricky122-5 

* Move `current_tx` up one level. 
* Keep `locator` in `fields`.
* CurrentTX traits like `FixedSizeFieldType` will go away, and be replaced by a Locator.

xrpl-common-stdlib/src/host/trace.rs
sappenin: Captured this idea in https://github.com/ripple/xrpl-wasm-stdlib/issues/238

Resolving.

sappenin (approved):

PR #214, opened 2026-07-09
feat: FieldDecoder Getters implemented and deployed for Current TX

ricky122-5:
### Type of Change

- [ ] Documentation update
- [ ] Release


xrpl-common-stdlib/src/fields/current_tx.rs
```diff
@@ -0,0 +1,309 @@
...
+use crate::fields::decoder::{FieldDecoder, FromCurrentTx, FromLedger};
+use crate::host;
+use crate::host::error_codes::FIELD_NOT_FOUND;
+use crate::host::{Result, get_tx_field};
+use crate::sfield::SField;
+use crate::types::decode_error::DecodeError;
+
+// `FieldDecoder` for the fixed-width unsigned integers the host writes directly. Each reads
+// exactly its own width and reinterprets the raw bytes as the host laid them out (native-endian
+// on wasm32, hence `from_ne_bytes`). All four are readable from both a transaction and a ledger
```
mvadari: I believe the endianness is fixed on host functions, so this shouldn't be necessary

xrpl-common-stdlib/src/fields/current_tx.rs
```diff
@@ -0,0 +1,309 @@
...
+// object.
+
+impl FieldDecoder for u8 {
+    type Buffer = [u8; 1];
```
mvadari, line 35: Why make this an array instead of just a `u8`?

xrpl-common-stdlib/src/fields/mod.rs
```diff
@@ -1,4 +1,12 @@
-//! Field decoding traits and helpers shared across XRPL field types.
+//! Typed accessors for reading fields from the current transaction and ledger objects, plus the
+//! [`locator::Locator`] builder for nested field paths.
+//!
+//! - [`decoder`]: Context-independent `FieldDecoder` trait + per-context marker traits
+//! - [`current_tx`]: Read fields from the current transaction
+//! - [`ledger_obj`]: Read fields from a ledger object by slot
+//! - [`locator`]: Build locators for nested field access
```
mvadari: Not necessary and hard to keep updated

xrpl-common-stdlib/src/objects/traits.rs
```diff
@@ -959,7 +959,6 @@ mod tests {
             assert_eq!(result.err().unwrap().code(), INVALID_FIELD);
         }
     }
-
```
mvadari, line 962: Why is this line removed? Seems better to keep it in

xrpl-common-stdlib/src/current_tx/traits.rs
```diff
@@ -452,17 +452,12 @@ mod tests {
                     .times(1)
                     .returning(|_, _, _| 0);
 
-                // Mock trace_num calls (2 calls per field for byte mismatch: expected + actual)
-                mock.expect_trace_num()
```
mvadari, line 456: Why are these tests removed? What changed that makes these tests not valid anymore?

xrpl-common-stdlib/src/fields/current_tx.rs
ricky122-5:
```rust
pub trait FieldDecoder: Sized {
    /// The buffer a `get_field` caller allocates before invoking the host function. Each type
    /// picks its own size (an associated type, not a `const`, so this stays on stable Rust).
    type Buffer: AsMut<[u8]>;

    /// Returns a zero-initialized buffer of this type's `Buffer` size.
    fn empty_buffer() -> Self::Buffer;

    /// Decodes `Self` from exactly the bytes the host wrote (not the full, possibly
    /// zero-padded, `Buffer`).
    fn decode(bytes: &[u8]) -> Result<Self, DecodeError>;
}
```

so the Buffer must be a u8 array

xrpl-common-stdlib/src/fields/current_tx.rs
ricky122-5: removed

xrpl-common-stdlib/src/objects/traits.rs
ricky122-5: Added back

xrpl-common-stdlib/src/fields/mod.rs
ricky122-5: Removed the member list

xrpl-common-stdlib/src/fields/current_tx.rs
```diff
@@ -0,0 +1,304 @@
...
+///
+/// // Type is automatically inferred from the SField constant
+/// let sequence = get_field(sfield::Sequence).unwrap();  // u32
+/// let account = get_field(sfield::Account).unwrap();  // AccountID
+/// ```
+#[inline]
+pub fn get_field<T: FromCurrentTx, const CODE: i32>(_: SField<T, CODE>) -> Result<T> {
+    let mut buf = T::empty_buffer();
+    let n = {
+        let slice = buf.as_mut();
```
mvadari, line 132: `buf` is already `mut`, why do we need to cast it again?

xrpl-common-stdlib/src/fields/decoder.rs
```diff
@@ -1,55 +1,32 @@
...
+use crate::types::decode_error::DecodeError;
 
-/// Decodes a fixed-format XRPL field from its raw byte representation.
+/// Decodes a typed value from the raw bytes a host function wrote.
 pub trait FieldDecoder: Sized {
-    /// A stack buffer sized to hold this field's raw bytes, used by generic host-field
-    /// getters to read into before calling [`decode`](FieldDecoder::decode).
-    type Buffer: AsMut<[u8]> + Default;
+    /// The buffer a `get_field` caller allocates before invoking the host function. Each type
+    /// picks its own size (an associated type, not a `const`, so this stays on stable Rust).
```
mvadari, line 15: This needs a TODO for why "stable Rust" matters and when we can migrate

xrpl-common-stdlib/src/current_tx/mod.rs
```diff
@@ -1,498 +1,3 @@
...
-/// ```
-#[inline]
-pub fn get_field_optional<T: CurrentTxFieldGetter, const CODE: i32>(
-    field: SField<T, CODE>,
-) -> Result<Option<T>> {
-    T::get_from_current_tx_optional(field)
-}
-
-#[cfg(test)]
-mod tests {
```
mvadari, line 278: Why are all these tests removed?

xrpl-common-stdlib/src/fields/current_tx.rs
```diff
@@ -0,0 +1,304 @@
...
+    type Buffer = [u8; 1];
+
+    #[inline]
+    fn empty_buffer() -> Self::Buffer {
+        [0u8; 1]
+    }
+
+    #[inline]
+    fn decode(bytes: &[u8]) -> core::result::Result<Self, DecodeError> {
+        let array: [u8; 1] = bytes.try_into().map_err(|_| DecodeError)?;
```
mvadari: Less repetitive this way:
```suggestion
        let array: Buffer = bytes.try_into().map_err(|_| DecodeError)?;
```
Ditto with the others

xrpl-escrow-stdlib/src/lib.rs
```diff
@@ -1,4 +1,21 @@
-#![doc = include_str!("../README.md")]
+//! # xrpl-escrow-stdlib
```
mvadari: These are unrelated changes and shouldn't be in this PR (it's already big enough)

xrpl-escrow-stdlib/src/current_tx/traits.rs
```diff
@@ -328,33 +328,41 @@ mod tests {
                 assert!(escrow.get_offer_sequence().is_ok());
             }
 
-            // Zero length for a mandatory fixed-size field panics (byte mismatch). One test
-            // per field, since `#[should_panic]` only catches the first panic.
+            // A zero-length read of a mandatory fixed-size field fails `FieldDecoder::decode`'s
```
mvadari, line 331: These are only semi-related changes and shouldn't be in this PR (it's already big enough

xrpl-common-stdlib/src/fields/decoder.rs
```diff
@@ -1,55 +1,32 @@
...
+    type Buffer: AsMut<[u8]>;
 
-    /// Decodes `Self` from `bytes`, returning an error if the bytes are malformed.
-    fn decode(bytes: &[u8]) -> Result<Self, Error>;
+    /// Returns a zero-initialized buffer of this type's `Buffer` size.
+    fn empty_buffer() -> Self::Buffer;
```
mvadari, line 22: I'd suggest making this addition a separate PR, because there's a number of changes that are related to just this change that aren't a result of the other factors.

mvadari (changes requested):
Is there a way to make this change piecemeal? This is a lot of code to review all at once, and most of it is a migration, not just moving code around

xrpl-common-stdlib/src/current_tx/traits.rs
ricky122-5: we removed trace_num() from decoding logic. it was previously embedded into decoding logic

xrpl-common-stdlib/src/fields/current_tx.rs
ricky122-5: It doesn't let me index into it directly bc it only satisfies AsMut. I could add AsRef, would make it more robust

xrpl-common-stdlib/src/fields/decoder.rs
ricky122-5: added

xrpl-common-stdlib/src/fields/decoder.rs
ricky122-5: empty_buffer is needed for the FieldDecoder to be functional (Buffer is its associated type)
we don't have T::Size bc const generic exprs isn't stable, so we are blind to buffer size in get_field. empty_buffer() allocates it for us using the impl'd FieldDecoder for that type

xrpl-common-stdlib/src/current_tx/mod.rs
ricky122-5: The traits don't exist anymore, these were tests for current tx getter that was replaced

xrpl-common-stdlib/src/fields/current_tx.rs
ricky122-5: Thanks, will update

xrpl-common-stdlib/src/fields/current_tx.rs
ricky122-5: will add AsRef for clarity

xrpl-escrow-stdlib/src/current_tx/traits.rs
ricky122-5: these are needed bc fixed size field accessors panicked on zero length reads, and this changes that by expecting the Err(InvalidDecoding) instead of the panic

xrpl-common-stdlib/src/fields/decoder.rs
mvadari: Yes, merge that PR before this one

xrpl-common-stdlib/src/current_tx/mod.rs
mvadari: Shouldn't the tests be migrated, not deleted?

xrpl-common-stdlib/src/current_tx/traits.rs
mvadari: Why?

xrpl-escrow-stdlib/src/lib.rs
ricky122-5: removed

xrpl-common-stdlib/src/current_tx/traits.rs
ricky122-5: My opinion: didn't want the side effect for decoding logic (what if someone didn't want to spend gas on that / don't want those logs), usually trace_num() is a function meant for users, and tests were hardcoded to constants (expected a certain amount of trace_num() calls that were hardcoded). Overall thought the cleaner approach was removal. Open to discussion for it though

xrpl-common-stdlib/src/current_tx/mod.rs
ricky122-5: We test the new FieldDecoder, yes. So I guess it's sort of a migration. But FieldDecoder != CurrentTxObjectGetter although they are similar

xrpl-common-stdlib/src/fields/current_tx.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    #[inline]
+    fn empty_buffer() -> Self::Buffer {
+        [0u8; 1]
+    }
+
+    #[inline]
+    fn decode(bytes: &[u8]) -> core::result::Result<Self, DecodeError> {
+        let array: Self::Buffer = bytes.try_into().map_err(|_| DecodeError)?;
+        Ok(u8::from_ne_bytes(array))
```
sappenin: This is technically incorrect. This code will generally be executed in WASM, which is always Little Endian. However, that's a bit of a red herring because these bytes will be coming from xrpld in XRPL's binary format (`ripple-binary-codec`) which always serializes integer fields (`UInt8`/`UInt16`/`UInt32`/`UInt64`) using big-endian.

In `FieldDecoder::decode` for `u16`, `u32`, and `u64` (currently using `from_ne_bytes`), we're decoding raw bytes copied straight off the wire from rippled via `get_tx_field` — no conversion happens in between. `get_tx_field` just memcpy's the serialized field bytes from the host into our buffer (and we expect those to be BE).

Since our WASM target is little-endian, `from_ne_bytes` here is effectively `from_le_bytes` — the opposite of what the bytes actually are. So today, decoding e.g. a `u32` SField will silently produce a wrong (byte-order-reversed) value in production WASM builds. The same is true for native unit tests: x86_64 and Apple Silicon are both little-endian, so `from_ne_bytes` behaves identically there — a test asserting a real BE fixture decodes correctly will catch this on any real machine, not just wasm.

For confirmation, everywhere else in the codebase that decodes a multi-byte value from a host buffer already uses `from_be_bytes` (see `types/nft.rs`, `types/mpt_id.rs`, `types/number.rs`, `types/amount.rs`, `types/opaque_float.rs` — the last one even has a comment: *"ALWAYS use `from_be_bytes` because that's how things are assembled"*).

Ironically, the `u8` decode is unaffected — with a single byte there's no ordering to get wrong, so `from_ne_bytes`/`from_be_bytes`/`from_le_bytes` are all equivalent here. It's only actually incorrect for the multi-byte cases.

Fix: change lines 63, 81, 99 (`u16`/`u32`/`u64`) from `from_ne_bytes` to `from_be_bytes`. We should always use `from_be_bytes` for binary-codec values regardless of target — it's not a wasm-vs-native distinction, it's just what the wire format is. We should also add a test that decodes a known big-endian byte fixture into its expected numeric value so a future refactor can't silently reintroduce this.

Separately worth a follow-up PR: `host/chain.rs` has what looks like the same bug in reverse — `from_le_bytes` for `ledger_sqn`/`parent_ledger_time`/`base_fee`, while the e2e test for those fields decodes with `from_be_bytes`.

xrpl-common-stdlib/src/fields/current_tx.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    #[inline]
+    fn empty_buffer() -> Self::Buffer {
+        [0u8; 2]
+    }
+
+    #[inline]
+    fn decode(bytes: &[u8]) -> core::result::Result<Self, DecodeError> {
+        let array: Self::Buffer = bytes.try_into().map_err(|_| DecodeError)?;
+        Ok(u16::from_ne_bytes(array))
```
sappenin:
```suggestion
        Ok(u16::from_le_bytes(array))
```

xrpl-common-stdlib/src/fields/current_tx.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    #[inline]
+    fn empty_buffer() -> Self::Buffer {
+        [0u8; 8]
+    }
+
+    #[inline]
+    fn decode(bytes: &[u8]) -> core::result::Result<Self, DecodeError> {
+        let array: Self::Buffer = bytes.try_into().map_err(|_| DecodeError)?;
+        Ok(u64::from_ne_bytes(array))
```
sappenin:
```suggestion
        Ok(u64:: from_le_bytes(array))
```

xrpl-common-stdlib/src/fields/current_tx.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    #[inline]
+    fn empty_buffer() -> Self::Buffer {
+        [0u8; 4]
+    }
+
+    #[inline]
+    fn decode(bytes: &[u8]) -> core::result::Result<Self, DecodeError> {
+        let array: Self::Buffer = bytes.try_into().map_err(|_| DecodeError)?;
+        Ok(u32::from_ne_bytes(array))
```
sappenin:
```suggestion
        Ok(u32:: from_le_bytes(array))
```

sappenin (changes requested):

xrpl-common-stdlib/src/fields/current_tx.rs
sappenin: Ok, thanks @ricky122-5 - my explanation is above is actually incorrect. `xrpld` is not using binary-codec to represent integer-based values. Instead, `xrpld` is adjusting from the host's native endianness to WASM's endianness (LE). On a LE host there's nothing to adjust; on a hypothetical BE host it byteswaps via `adjustWasmEndianess`. So, the output is always LE in the WASM buffer.for byte values.

I'll reverse my suggestions below.

xrpl-common-stdlib/src/types/amount.rs
```diff
@@ -338,39 +339,29 @@ impl LedgerObjectFieldGetter for Amount {
...
-        field: SField<Self, CODE>,
-    ) -> Result<Option<Self>> {
-        get_variable_size_field_optional::<AMOUNT_SIZE, _>(
-            i32::from(field),
-            |fc, buf, size| unsafe { get_tx_field(fc, buf, size) },
-        )
-        .map(|opt| opt.map(|(buffer, _len)| Amount::from(buffer)))
+    fn decode(bytes: &[u8]) -> core::result::Result<Self, DecodeError> {
+        let mut padded = [0u8; AMOUNT_SIZE];
+        let n = bytes.len().min(AMOUNT_SIZE);
```
sappenin, line 356: Clarify the assumption here on `bytes`: I'm under the impression that when an `STAmount`comes from the host, it always has a byte-length of 48 bytes no matter what. When parsing to a Rust `Amount`, the leading bits will determine the actual length to parse (8, 33, 48) depending on token type.

If that's true, and given `Amount::from_bytes` isn't called until later (line `358) then I don't think clamping to `AMOUNT_SIZE` (i.e., 48) bytes is the correct behavior. Rationale is that the `bytes` buffer should _always_ be 48 bytes, no matter what. If this function receives a buffer with more or less-than 48 bytes, it should throw a decode error (like we do for `FieldDecoder` for `AccountID`). 

So, I have a few assumptions above....but if they're all true, I think this function should instead be: 

```rust
    let padded: [u8; AMOUNT_SIZE] = bytes.try_into().map_err(|_| DecodeError)?;
    Amount::from_bytes(&padded).ok().ok_or(DecodeError)
}
```

sappenin (changes requested):

xrpl-common-stdlib/src/types/amount.rs
ricky122-5: I think the first assumption may not be true, here's get_field from xrpl-common-stdlib/src/fields/current_tx.rs:129-149:
```rust
pub fn get_field<T: FromCurrentTx, const CODE: i32>(_: SField<T, CODE>) -> Result<T> {
    let mut buf = T::empty_buffer();
    let n = {
        let slice = buf.as_mut();
        unsafe { get_tx_field(CODE, slice.as_mut_ptr(), slice.len()) }
    };
    if n < 0 {
        return Result::Err(host::Error::from_code(n));
    }
    let bytes = buf.as_ref();
    let n = n as usize;
    if n > bytes.len() {
        return Result::Err(host::Error::PointerOutOfBounds);
    }
    match T::decode(&bytes[..n]) {   // <-- sliced to n, the host's reported length
        core::result::Result::Ok(value) => Result::Ok(value),
        core::result::Result::Err(_) => Result::Err(host::Error::InvalidDecoding),
    }
}
```
`buf` is still 48 bytes but `n` is the actual # of bytes written, and only the `&bytes[..n] goes to `Amount::decode`. For current_tx, I think this should be expected behavior

PR #226, opened 2026-07-15
feat: Add contract-only host function bindings (data storage, txn, etc.)

ricky122-5:

Adds Contract-specific host function bindings

### Type of Change

- [ ] Documentation update
- [ ] Release

…t, params) to HostBindings trait


tools/compareHostFunctions.js
```diff
@@ -20,6 +20,32 @@ function areListsEqual(a, b) {
...
 
+// Host functions that exist in the Rust stdlib ahead of the C++ rippled
+// implementation. The audit tolerates these being "extra in Rust" so the PR
+// can land without blocking on the rippled side. Remove a name from this list
+// once the corresponding C++ host function is merged into rippled.
+//
+// All entries below are smart-contract entry-point/data/event/transaction
+// builder functions that have not yet been specified or implemented in
+// rippled's WasmVM.cpp / HostFuncWrapper.h.
+const PENDING_CPP_ALLOWLIST = new Set([
```
mvadari, line 31: This is hard to maintain if other host functions are added here

ricky122-5: This was in the contract/ branch. I think it's necessary to pass CI as we don't have any xrpld/ implementations but I'm open to making it contract-only / naming it to indicate that, or maybe going for another solution

sappenin: Can we pull Smart Contract changes into their own temporary parent branch without pushing them to `main`? Smart Contracts aren't real yet, and I'd prefer to finish all the "moving around" stuff first rather than trying to simultaneously "move around" + "add new primitive."

Thoughts?

xrpl-wasm-stdlib/src/host/host_bindings_trait.rs
```diff
@@ -1643,4 +1643,465 @@ pub trait HostBindings {
...
+    ///
+    /// # Returns
+    ///
+    /// - Returns a positive number of bytes written to the output buffer on success
+    /// - Returns a negative error code on failure. The list of error codes is defined in
+    ///   `../core/error_codes.rs`
+    ///
+    /// # Safety
+    /// Caller must ensure all pointer parameters point to valid memory
+    unsafe fn instance_param(
```
sappenin, line 1671: I don't think this is the correct place for Smart Contract-related Host Function definitions. My rationale is I'd like to keep a clear distinction between host functions available to a Smart Escrow vs. a Smart Contract (i.e., a Smart Escrow developer shouldn't see Smart Contract trait functions, but this PR conflates the two).

Instead, I think we should use different traits here. We can keep the `HostBindings` trait for anything related to a Smart Escrow. Then, we can create a new traits that extends this, e.g.:

```rust
pub trait SmartContractHostBindings : HostBindings { ...}
```

For a developer using the SDK, this will keep the trait functions clean and won't allow a Smart Escrow developer to accidentally invoke something that doesn't exist. 

Thoughts?

sappenin (changes requested):

PR #232, opened 2026-07-15
feat: Inject version information into wasm module

TimothyBanks:
### Type of Change

- [ ] Documentation update
- [ ] Release


xrpl-common-stdlib/src/lib.rs
```diff
@@ -4,6 +4,16 @@
 #[cfg(not(target_arch = "wasm32"))]
 extern crate std;
 
+#[cfg(target_arch = "wasm32")]
+const VERSION_STR: &str = env!("CARGO_PKG_VERSION");
+
+#[cfg(target_arch = "wasm32")]
+#[used]
+#[unsafe(link_section = "xrpl-common-stdlib-version")]
+static VERSION_METADATA: [u8; VERSION_STR.len()] = *unsafe {
```
TimothyBanks: For some reason the injection only works consistently with release builds.

Also, I am not proposing that we use these version numbers.  It is just an example that we could embed version information when compiling the wasm module.

PR #233, opened 2026-07-15
feat: Crypto host function wrappers

ricky122-5:

Adds wrappers for host functions:
**`sha512_half` contract** (`HostFuncWrapper.cpp:848`, `HostFuncImpl.cpp:47`):
- Input: variable-length data, 0 to `maxWasmParamLength` (1024) bytes
- Output: 32-byte SHA-512-half hash written to caller-supplied buffer
- Errors: `DataFieldTooLarge (-12)` if input > 1024 bytes

**`check_sig` contract** (`HostFuncImpl.cpp:35`, `PublicKey.cpp:270`, `HostFuncWrapper.cpp:817`):
- Parameters: `message` (0-1024 bytes), `sig` (0-1024 bytes), `pubkey` (0-1024 bytes)
- Returns `1` (valid), `0` (invalid), or negative error code
- `InvalidParams (-15)`: pubkey fails `publicKeyType()` -- not exactly 33 bytes or first byte not in `{0xED, 0x02, 0x03}`
- `DataFieldTooLarge (-12)`: any parameter > 1024 bytes
- `0`: empty message or empty sig is not an error -- verify returns false
- For secp256k1 keys (`0x02`/`0x03`): rippled pre-hashes the message with sha512Half before ECDSA verify
- For Ed25519 keys (`0xED`): rippled verifies directly on the raw message (no pre-hash); strips the `0xED` prefix byte before calling ed25519_sign_open; non-canonical sig returns `false`
- `PublicKey` in the stdlib is `[u8; 33]` -- use `&PublicKey` rather than `&[u8]` to enforce the size constraint at the call site

### Type of Change

- [ ] Documentation update
- [ ] Release

### DIFF
--- a/xrpl-common-stdlib/src/crypto.rs
+++ b/xrpl-common-stdlib/src/crypto.rs
@@ -78,6 +78,18 @@ mod tests {
     }
 
     #[test]
+    #[should_panic(expected = "internal invariant violated")]
+    fn test_sha512_half_wrong_byte_count() {
+        let mut mock = MockHostBindings::new();
+        mock.expect_compute_sha512_half()
+            .times(1)
+            .returning(|_, _, _, _| 16); // host returns wrong (non-32) byte count
+        let _guard = setup_mock(mock);
+
+        let _ = sha512_half(b"hello");
+    }
+
+    #[test]
     fn test_sha512_half_oversized() {
         let mut mock = MockHostBindings::new();
         mock.expect_compute_sha512_half()
--- a/xrpl-common-stdlib/src/crypto.rs
+++ b/xrpl-common-stdlib/src/crypto.rs
@@ -25,8 +25,7 @@ pub fn sha512_half(data: &[u8]) -> Result<[u8; 32]> {
 ///
 /// Errors: `InvalidParams` if the key is malformed (not 33 bytes / bad prefix);
 /// `DataFieldTooLarge` if any parameter exceeds 1024 bytes.
-pub fn check_sig(key: &PublicKey, msg: &[u8], sig: &[u8]) -> Result<bool> {
-    // Host arg order is (message, signature, pubkey); the wrapper reorders to (key, msg, sig).
+pub fn check_sig(msg: &[u8], sig: &[u8], key: &PublicKey) -> Result<bool> {
     let rescode = unsafe {
         host::check_sig(
             msg.as_ptr(),
@@ -120,7 +119,7 @@ mod tests {
         let _guard = setup_mock(mock);
 
         let key = PublicKey::from(PUBKEY_BYTES);
-        let result = check_sig(&key, b"message", b"signature");
+        let result = check_sig(b"message", b"signature", &key);
         assert!(result.unwrap());
     }
 
@@ -133,7 +132,7 @@ mod tests {
         let _guard = setup_mock(mock);
 
         let key = PublicKey::from(PUBKEY_BYTES);
-        let result = check_sig(&key, b"message", b"signature");
+        let result = check_sig(b"message", b"signature", &key);
         assert!(!result.unwrap());
     }
 
@@ -146,7 +145,7 @@ mod tests {
         let _guard = setup_mock(mock);
 
         let key = PublicKey::from(PUBKEY_BYTES);
-        let result = check_sig(&key, b"message", b"signature");
+        let result = check_sig(b"message", b"signature", &key);
         assert!(result.is_err());
         assert_eq!(result.err().unwrap().code(), INVALID_PARAMS);
     }
@@ -160,7 +159,7 @@ mod tests {
         let _guard = setup_mock(mock);
 
         let key = PublicKey::from(PUBKEY_BYTES);
-        let result = check_sig(&key, &[0u8; 1025], b"signature");
+        let result = check_sig(&[0u8; 1025], b"signature", &key);
         assert!(result.is_err());
         assert_eq!(result.err().unwrap().code(), DATA_FIELD_TOO_LARGE);
     }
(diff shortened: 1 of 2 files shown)


xrpl-common-stdlib/src/crypto.rs
```diff
@@ -0,0 +1,154 @@
...
+                32
+            });
+        let _guard = setup_mock(mock);
+
+        let result = sha512_half(b"hello world");
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap(), [0xCD; 32]);
+    }
+
+    #[test]
```
mvadari, line 92: There should be a test to ensure it errors properly if the return value is not 32 - we want to check primarily for errors that the stdlib checks for, not rippled

ricky122-5: added

xrpl-common-stdlib/src/crypto.rs
```diff
@@ -0,0 +1,166 @@
...
+            sig.as_ptr(),
+            sig.len(),
+            key.0.as_ptr(),
+            PUBLIC_KEY_BUFFER_SIZE,
+        )
+    };
+    match rescode {
+        0 => Result::Ok(false),
+        1 => Result::Ok(true),
+        _ => Result::Err(Error::from_code(rescode)),
```
mvadari: Only a negative value should be turned into an error, a different positive value should panic

xrpl-common-stdlib/src/crypto.rs
```diff
@@ -0,0 +1,166 @@
...
+/// `&PublicKey` (33 bytes) enforces the size constraint at the call site.
+/// - secp256k1 keys (0x02/0x03): the host pre-hashes `msg` with SHA-512Half before ECDSA verify.
+/// - Ed25519 keys (0xED): the host verifies the raw `msg` directly (no pre-hash), stripping the
+///   0xED prefix; a non-canonical signature returns `Ok(false)`.
+///
+/// An empty `msg` or empty `sig` is not an error — the host returns `Ok(false)`.
+///
+/// Errors: `InvalidParams` if the key is malformed (not 33 bytes / bad prefix);
+/// `DataFieldTooLarge` if any parameter exceeds 1024 bytes.
+pub fn check_sig(key: &PublicKey, msg: &[u8], sig: &[u8]) -> Result<bool> {
```
kuznetsss: Shouldn't we introduce message and signature types? With the current function signature it is easy to put args in wrong order `check_sig(key, sig, msg)`. But with custom types such mistake would be caught at compile time

xrpl-common-stdlib/src/crypto.rs
ricky122-5: fixed

xrpl-common-stdlib/src/crypto.rs
mvadari: I assume just basically alias types? Wouldn't you figure that out with one quick test?

xrpl-common-stdlib/src/crypto.rs
```diff
@@ -0,0 +1,180 @@
...
+    // Host arg order is (message, signature, pubkey); the wrapper reorders to (key, msg, sig).
```
mvadari: Why change the order?

ricky122-5: I feel like it reads natural for a user to see the function and go for this key, verify this (msg, sig) pair rather than sig and then msg. if we want to strictly match the ABI that's fine

mvadari: We can change the ABI if we want, but the helper function should be the same as the host function.

ricky122-5: fixed, ty

xrpl-common-stdlib/src/crypto.rs
```diff
@@ -25,8 +25,7 @@ pub fn sha512_half(data: &[u8]) -> Result<[u8; 32]> {
 ///
 /// Errors: `InvalidParams` if the key is malformed (not 33 bytes / bad prefix);
 /// `DataFieldTooLarge` if any parameter exceeds 1024 bytes.
-pub fn check_sig(key: &PublicKey, msg: &[u8], sig: &[u8]) -> Result<bool> {
-    // Host arg order is (message, signature, pubkey); the wrapper reorders to (key, msg, sig).
+pub fn check_sig(key: &PublicKey, sig: &[u8], msg: &[u8]) -> Result<bool> {
```
mvadari: Should be msg, sig, key to match the host function

ricky122-5: yea mb, fixing

xrpl-common-stdlib/src/crypto.rs
ricky122-5: fixed

mvadari (approved):

xrpl-common-stdlib/src/crypto.rs
sappenin: I agree with @kuznetsss that we should introduce message and signature types. However, I would be fine just tracking that in an issue and doing it in a different PR. 

@kuznetsss any objection to that, to unblock this one?

sappenin (approved):

xrpl-common-stdlib/src/crypto.rs
kuznetsss: Issue and separate PR to fix it sounds good. I guess this is not the only place in the code to fix.

> I assume just basically alias types?

Not alias, but a new type, something like:
```rust
struct Signature([u8; 8]);
```

> Wouldn't you figure that out with one quick test?

It is more convenient to catch errors at compile time. New type is a standard recommended pattern: https://doc.rust-lang.org/rust-by-example/generics/new_types.html

kuznetsss (approved):

PR #234, opened 2026-07-16
fix: Update sfields custom types to include Signature

ricky122-5:

Removes hardcoding on sfields.rs of Signature and adds it to the customTypesList

### Type of Change

- [ ] Documentation update
- [ ] Release


examples/Cargo.lock
kuznetsss, line 20: It seems changes in `Cargo.lock` are out of scope of this PR. Please update it in a separate PR if needed

ricky122-5: I believe that's autogenerated, do we need another pr?

mvadari: Dependency updates should be separate. You never know when a dependency update will mess up something, so better to be safe rather than sorry.

xrpl-wasm-stdlib/src/sfield.rs
```diff
@@ -414,6 +414,8 @@ pub const NFTokenOffers: SField<u8, 1245188> = SField::new();
 pub const CredentialIDs: SField<u8, 1245189> = SField::new();
 pub const MPTokenIssuanceID: SField<Hash192, 1376257> = SField::new();
 pub const ShareMPTID: SField<Hash192, 1376258> = SField::new();
+pub const TakerPaysMPT: SField<Hash192, 1376259> = SField::new();
+pub const TakerGetsMPT: SField<Hash192, 1376260> = SField::new();
```
mvadari, line 418: Where did these come from? And shouldn't `Signature` have changed?

sappenin (approved):

xrpl-wasm-stdlib/src/sfield.rs
ricky122-5: They might just be in the branch bc these were autogenerated. Signature's sfield was hardcoded before which this fixed.

examples/Cargo.lock
ricky122-5: I think David was able to remove it from the diff. Thanks!

mvadari (approved):

PR #241, opened 2026-07-27
feat: Migrate types to FieldDecoder, empty_buffer()

ricky122-5:

Split of #214 to contain only the diffs for empty_buffer() addition to FieldDecoder and accompanying migrations on types. The following PR will contain the next stage of the split.

### Type of Change

- [ ] Documentation update
- [ ] Release

### DIFF
--- a/xrpl-common-stdlib/src/types/amount.rs
+++ b/xrpl-common-stdlib/src/types/amount.rs
@@ -386,10 +386,8 @@ impl FieldDecoder for Amount {
 
     #[inline]
     fn decode(bytes: &[u8]) -> core::result::Result<Self, DecodeError> {
-        let mut padded = [0u8; AMOUNT_SIZE];
-        let n = bytes.len().min(AMOUNT_SIZE);
-        padded[..n].copy_from_slice(&bytes[..n]);
-        Amount::from_bytes(&padded).ok().ok_or(DecodeError)
+        let amount_bytes: [u8; AMOUNT_SIZE] = bytes.try_into().map_err(|_| DecodeError)?;
+        Amount::from_bytes(&amount_bytes).ok().ok_or(DecodeError)
     }
 }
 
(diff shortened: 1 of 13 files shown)


xrpl-common-stdlib/src/types/amount.rs
```diff
@@ -371,6 +373,29 @@ impl CurrentTxFieldGetter for Amount {
...
+impl FieldDecoder for Amount {
+    type Buffer = [u8; AMOUNT_SIZE];
+
+    #[inline]
+    fn empty_buffer() -> Self::Buffer {
+        [0u8; AMOUNT_SIZE]
+    }
+
+    #[inline]
+    fn decode(bytes: &[u8]) -> core::result::Result<Self, DecodeError> {
```
sappenin, line 388: Continuing the discussion here from [this discussion in #214](https://github.com/ripple/xrpl-wasm-stdlib/pull/214#discussion_r3657760928).

In your example/comment from #214, the variable `n` should always be 48 bytes because the host will always return an `STAmount` back using 48 bytes. It's not until `Amount::from_bytes` that the actual 8/33/48 bytes are parsed. 

For supporting evidence, notice the invariant on line 195 of `amount.rs` (i.e., if the bytes sent into `Amount::from_bytes` are ever _not_ 48 bytes, an error will be thrown). Also see the unit test around line 540.

I point this out because `n` below (in this file) will always be 48 bytes, so there's no point in checking it here. Instead, this can be simplified to this:

```rust
    let amount_bytes: [u8; AMOUNT_SIZE] = bytes.try_into().map_err(|_| DecodeError)?;
    Amount::from_bytes(&amount_bytes).ok().ok_or(DecodeError)
}
```

sappenin (changes requested):

xrpl-common-stdlib/src/types/amount.rs
ricky122-5: Makes sense, will make the change. Thanks!

sappenin (approved):

PR #242, opened 2026-07-27
feat: Implement getters for FieldDecoder and migrate types

ricky122-5:

Implements the getters for FieldDecoder and removes old getters from types, wiring in the new FieldDecoder.

### Type of Change

- [ ] Documentation update
- [ ] Release


xrpl-common-stdlib/src/fields/current_tx.rs
```diff
@@ -0,0 +1,323 @@
...
+//! companion crates (`xrpl-escrow-stdlib` for escrow flows).
+
+use crate::fields::decoder::{FieldDecoder, FromCurrentTx, FromLedger};
+use crate::host;
+use crate::host::error_codes::FIELD_NOT_FOUND;
+use crate::host::{Result, get_tx_field};
+use crate::sfield::SField;
+use crate::types::decode_error::DecodeError;
+
+impl FieldDecoder for u8 {
```
sappenin: I'm not sure it's correct to put this trait impl in this file. Primitive unsigned integers can exist in more than just the current transaction, so putting their `FieldDecoder` impl in here seems off.  For example, I'm pretty sure unsigned primitive numbers can also can show up in a ledger object fields, right? 

If true, I would propose these impls land in one of the following more generic locations: `decoder.rs`, `uint.rs`, or maybe `primitives.rs`?

@mvadari do you have an opinion?

xrpl-common-stdlib/src/fields/decoder.rs
```diff
@@ -32,78 +32,6 @@ pub trait FromCurrentTx: FieldDecoder {}
 /// Marker: this type can be read from a ledger object via [`crate::fields::ledger_obj`].
 pub trait FromLedger: FieldDecoder {}
 
-impl FieldDecoder for u8 {
```
sappenin, line 35: See my comment in `current_tx.rs`

sappenin (changes requested):

xrpl-common-stdlib/src/fields/current_tx.rs
ricky122-5: moved to decoder.rs

sappenin (dismissed):

sappenin (approved):

xrpl-common-stdlib/src/current_tx/traits.rs
```diff
@@ -661,66 +644,77 @@ mod tests {
...
+            fn test_get_account_errors_when_zero_length() {
                 let mut mock = MockHostBindings::new();
                 mock.expect_get_tx_field()
                     .with(eq(sfield::Account), always(), eq(ACCOUNT_ID_SIZE))
                     .returning(|_, _, _| 0);
 
                 let _guard = setup_mock(mock);
-                let _ = TestTransaction.get_account();
+                let result = TestTransaction.get_account();
+                assert!(result.is_err());
```
TimothyBanks, line 664: The `!` throws me off every time as I read this as "assert not result is error".

xrpl-common-stdlib/src/fields/current_tx.rs
```diff
@@ -0,0 +1,250 @@
...
+/// ```rust,no_run
+/// use xrpl_common_stdlib::fields::current_tx::get_field;
+/// use xrpl_common_stdlib::sfield;
+///
+/// // Type is automatically inferred from the SField constant
+/// let sequence = get_field(sfield::Sequence).unwrap();  // u32
+/// let account = get_field(sfield::Account).unwrap();  // AccountID
+/// ```
+#[inline]
+pub fn get_field<T: FromCurrentTx, const CODE: i32>(_: SField<T, CODE>) -> Result<T> {
```
kuznetsss, line 54: Should `get_field()` reuse `get_field_optional()`? They look almost identical

xrpl-common-stdlib/src/fields/ledger_obj.rs
```diff
@@ -0,0 +1,204 @@
...
+use crate::sfield::SField;
+
+/// Retrieves a field from the ledger object cached in `slot` using an SField constant.
+///
+/// # Returns
+///
+/// Returns a `Result<T>` where:
+/// * `Ok(T)` - The field value for the specified field
+/// * `Err(Error)` - If the field cannot be retrieved, has unexpected size, or fails to decode
+#[inline]
```
kuznetsss, line 19: Why should it be annotated as `inline`?

xrpl-common-stdlib/src/fields/current_tx.rs
```diff
@@ -0,0 +1,250 @@
...
+///
+/// // Type is automatically inferred from the SField constant
+/// let flags = get_field_optional(sfield::Flags).unwrap();  // Option<u32>
+/// let source_tag = get_field_optional(sfield::SourceTag).unwrap();  // Option<u32>
+/// ```
+#[inline]
+pub fn get_field_optional<T: FromCurrentTx, const CODE: i32>(
+    _: SField<T, CODE>,
+) -> Result<Option<T>> {
+    let mut buf = T::empty_buffer();
```
TimothyBanks: I'm not up to speed on Rust and code deduplication techniques within it, but for instance, it seems like this method could call `get_field` or vice versa.  Or maybe there is something generic programming idiom in Rust that could handle this as well.

xrpl-common-stdlib/src/fields/ledger_obj.rs
```diff
@@ -0,0 +1,204 @@
...
+pub fn get_field<T: FromLedger, const CODE: i32>(slot: i32, _: SField<T, CODE>) -> Result<T> {
+    let mut buf = T::empty_buffer();
```
TimothyBanks, line 21: Then building on my last comment about code deduplication, it seems like there is a pattern here with the `get_field` and `get_field_optional` implementations in general that would fit nicely into some sort of generic.

xrpl-common-stdlib/src/fields/current_tx.rs
TimothyBanks: I left the same comment later on, apologies for trying to steal your thunder here!

xrpl-common-stdlib/src/types/amount.rs
```diff
@@ -386,8 +352,14 @@ impl FieldDecoder for Amount {
 
     #[inline]
     fn decode(bytes: &[u8]) -> core::result::Result<Self, DecodeError> {
-        let amount_bytes: [u8; AMOUNT_SIZE] = bytes.try_into().map_err(|_| DecodeError)?;
```
ricky122-5, line 389: @sappenin reverted this back to the variable size logic. 

The reason behind this is because the host wasn't always returning STAmount's 48 bytes, but returned the amount of bytes written to the buffer itself. This caused a problem in integration tests with trace_escrow_finish throwing error code -11, trying to decode a Fee (8 byte XRP type) on get_fee(), where it would fail the try_into() statement.

However I made it a bit more robust to only accept valid Amount types. LMK thoughts

xrpl-common-stdlib/src/fields/ledger_obj.rs
ricky122-5: Just avoids a real call in the `no_std` WASM target. Optimization wrapper.

xrpl-common-stdlib/src/fields/ledger_obj.rs
ricky122-5: Thanks @kuznetsss @TimothyBanks ! Will make a fix

TimothyBanks (approved):
I like it!

xrpl-common-stdlib/src/fields/decoder.rs
```diff
@@ -32,6 +33,30 @@ pub trait FromCurrentTx: FieldDecoder {}
...
+        return host::Result::Err(host::Error::from_code(n));
+    }
+    let bytes = buf.as_ref();
+    let n = n as usize;
+    if n > bytes.len() {
+        // A conformant host never reports writing more bytes than the buffer holds; a positive
+        // count past our buffer means it described memory outside the allowed region.
+        return host::Result::Err(host::Error::PointerOutOfBounds);
+    }
+    match T::decode(&bytes[..n]) {
```
sappenin: This line is incorrect. It subtly truncating two pieces of information (capacity + length) into a single slice, which loses fidelity.

In the `::decode` trait, instead of a slice we should define this:

```rust
pub type AmountBuffer = Blob<AMOUNT_SIZE>;
```

and have it take an `AmountBuffer`.

xrpl-common-stdlib/src/types/amount.rs
```diff
@@ -386,8 +352,14 @@ impl FieldDecoder for Amount {
...
-        Amount::from_bytes(&amount_bytes).ok().ok_or(DecodeError)
+        match bytes.len() {
+            8 | 33 | AMOUNT_SIZE => {
```
sappenin: See my other comment. One reason we don't want to do this is just separation of concerns. We want `amount.rs` to know about the internal meaning of these bytes (not this trait impl). If we fix the Decoder trait, I think this will go away.

sappenin (changes requested):

xrpl-common-stdlib/src/fields/ledger_obj.rs
sappenin: Mostly as a hint. The compiler will decide if it should inline, but it doesn't have to if it decides not to. Also, this is a tiny helper function around an FFI call, so inlining is more efficient than call function overhead (every function call generates a bunch of WASM instruction, so inlining likely lowers WASM instruction counts by eliminating call and other instructions directly reducing the gas required). 

If memory serves, we ran some WASM benchmarks on these a while back, but I don't know that we captured the results anywhere.

Claude also tells me this, among other things:

> SROA (Scalar Replacement of Aggregates): If the caller immediately consumes the result T and breaks it down, inlining allows LLVM to completely eliminate buf from linear memory altogether, replacing it entirely with WASM local variables (i32/i64 registers).

xrpl-common-stdlib/src/types/transaction_type.rs
```diff
@@ -200,9 +166,11 @@ impl FieldDecoder for TransactionType {
...
 
     #[inline]
-    fn decode(bytes: &[u8]) -> core::result::Result<Self, DecodeError> {
-        let array: Self::Buffer = bytes.try_into().map_err(|_| DecodeError)?;
-        Ok(array.into())
+    fn decode(buf: &Self::Buffer, bytes_written: usize) -> core::result::Result<Self, DecodeError> {
+        if bytes_written != buf.len() {
+            return core::result::Result::Err(DecodeError);
+        }
+        Ok((*buf).into())
```
sappenin: I'm noticing that `FieldDecoder` is copying all buffers unnecessarily. See https://github.com/ripple/xrpl-wasm-stdlib/pull/247 (targeting this branch) for a solution to stop doing that.

sappenin (changes requested):

xrpl-common-stdlib/src/fields/ledger_obj.rs
kuznetsss: Yeah, `#[inline]` doesn't guarantee inlining, it is only adds a small shift to inlining threshold. Maybe we shouldn't put `#[inline]` everywhere without performance measurements? (Not a blocker for this PR)

xrpl-common-stdlib/src/fields/current_tx.rs
```diff
@@ -0,0 +1,225 @@
...
+    let n = {
+        let slice = buf.as_mut();
+        unsafe { get_tx_field(CODE, slice.as_mut_ptr(), slice.len()) }
+    };
+    if n == FIELD_NOT_FOUND {
+        return Result::Ok(None);
+    }
+    decode_result::<T>(&buf, n).map(Some)
+}
```
sappenin, line 95:
```suggestion
#[inline]
pub fn get_field_optional<T: FromCurrentTx, const CODE: i32>(
    field: SField<T, CODE>,
) -> Result<Option<T>> {
    match get_field(field) {
        Result::Ok(value) => Result::Ok(Some(value)),
        Result::Err(Error::FieldNotFound) => Result::Ok(None),
        Result::Err(e) => Result::Err(e),
    }
}
```

sappenin (changes requested):

xrpl-common-stdlib/src/types/amount.rs
```diff
@@ -385,9 +352,11 @@ impl FieldDecoder for Amount {
...
 
     #[inline]
-    fn decode(bytes: &[u8]) -> core::result::Result<Self, DecodeError> {
-        let amount_bytes: [u8; AMOUNT_SIZE] = bytes.try_into().map_err(|_| DecodeError)?;
-        Amount::from_bytes(&amount_bytes).ok().ok_or(DecodeError)
+    fn decode(buf: &Self::Buffer, bytes_written: usize) -> core::result::Result<Self, DecodeError> {
+        match bytes_written {
+            8 | 33 | AMOUNT_SIZE => Amount::from_bytes(buf).ok().ok_or(DecodeError),
+            _ => core::result::Result::Err(DecodeError),
+        }
```
sappenin, line 371:
```suggestion
#[inline]
    fn decode(buf: Self::Buffer, bytes_written: usize) -> core::result::Result<Self, DecodeError> {
        // Unlike `Issue`, `Amount`'s variant is self-describing via the flag bits in byte 0,
        // present regardless of how many bytes were written, so `bytes_written` can't be used to
        // *pick* the variant. But once the variant is known, XRPL's wire format fixes exactly how
        // many bytes a conformant host writes for it (8 XRP / 33 MPT / 48 IOU) — trusting a
        // `bytes_written` that's inconsistent with the parsed variant would mean silently
        // accepting a truncated or otherwise malformed host response as a valid (but wrong)
        // value, rather than surfacing it as a decode error.
        let amount = Amount::from(buf);
        let expected_len = match amount {
            Amount::XRP { .. } => 8,
            Amount::MPT { .. } => 33,
            Amount::IOU { .. } => AMOUNT_SIZE,
        };
        if bytes_written != expected_len {
            return core::result::Result::Err(DecodeError);
        }
        core::result::Result::Ok(amount)
    }
```

sappenin (changes requested):

xrpl-common-stdlib/src/fields/decoder.rs
```diff
@@ -21,9 +22,13 @@ pub trait FieldDecoder: Sized {
...
-    /// Decodes `Self` from exactly the bytes the host wrote (not the full, possibly
-    /// zero-padded, `Buffer`).
-    fn decode(bytes: &[u8]) -> Result<Self, DecodeError>;
+    /// Decodes `Self` from the full `Buffer` (as written by the host, then zero-padded to the
+    /// buffer's size), given `bytes_written` — the number of bytes the host actually wrote.
+    ///
+    /// `bytes_written` carries the length that a `&[u8]` slice would otherwise bundle inside its
+    /// fat pointer — passing the whole buffer plus `bytes_written` lets fixed-layout types (e.g.
+    /// `Amount`) read the padded buffer in place, with no re-slice or re-copy.
+    fn decode(buf: &Self::Buffer, bytes_written: usize) -> Result<Self, DecodeError>;
```
sappenin, line 38:
```suggestion
    ///
    /// Most fixed-size types require the host to have written *exactly* `Buffer`'s length and can
    /// then be built with a plain `From<Buffer>` — for those, implement this as
    /// `decode_exact(buf, bytes_written)` (see [`decode_exact`]). Types with different semantics
    /// — e.g. `Amount`, where a shorter write is legitimate (XRP is 8 bytes, MPT 33, of a 48-byte
    /// buffer) and the variant is determined by the leading flag bits rather than the byte count
    /// — write a bespoke body instead.
    fn decode(buf: &Self::Buffer, bytes_written: usize) -> Result<Self, DecodeError>;
```

xrpl-common-stdlib/src/fields/decoder.rs
```diff
@@ -21,9 +22,13 @@ pub trait FieldDecoder: Sized {
...
 }
```
sappenin, line 40:
```suggestion
/// Shared `FieldDecoder::decode` body for fixed-size types that require an exact-length write and
/// build `Self` via `From<Buffer>` (`AccountID`, `TransactionType`, `UInt<N>`). Not a default
/// trait method: a `where Self: From<Self::Buffer>` bound on a trait method applies to every
/// implementor, including ones that override the body, so a type without that `From` impl (e.g.
/// `Amount`, whose variant is decided by flag bits rather than length) couldn't compile at all.
/// A plain function sidesteps that — callers opt in explicitly.
#[inline]
pub(crate) fn decode_exact<T, Buf>(buf: Buf, bytes_written: usize) -> Result<T, DecodeError>
where
    Buf: AsRef<[u8]>,
    T: From<Buf>,
{
    if bytes_written != buf.as_ref().len() {
        return Err(DecodeError);
    }
    Ok(T::from(buf))
}

```

xrpl-common-stdlib/src/fields/decoder.rs
```diff
@@ -32,6 +37,29 @@ pub trait FromCurrentTx: FieldDecoder {}
...
 pub trait FromLedger: FieldDecoder {}
 
+/// Shared step behind every `get_field`/`get_field_optional` in [`crate::fields::current_tx`]
+/// and [`crate::fields::ledger_obj`]: turn a host result code and the buffer it (partially)
+/// filled into a typed value.
+///
+/// Callers handle the "field not found" case themselves (only `get_field_optional` has one)
+/// before reaching here; `n` is assumed to be either a real byte count or a hard error.
+#[inline]
+pub(crate) fn decode_result<T: FieldDecoder>(buf: &T::Buffer, n: i32) -> host::Result<T> {
```
sappenin: AI flagged the naming here. 

> There's a real problem: decode_result, and FieldDecoder::decode share the decode prefix despite being different things (driver, shared-body helper, trait method), so a reader has to check signatures to know which layer they're looking at. `decode_result` is the worse offender since its name doesn't hint at what makes it distinct — that it's the one consuming the host's raw i32 code. I'd rename it to something like decode_host_result (or resolve_host_result) to make that explicit.

```suggestion
pub(crate) fn decode_host_result<T: FieldDecoder>(buf: &T::Buffer, n: i32) -> host::Result<T> {
```

sappenin (approved):

Issue #116, opened 2025-12-11
Update struct accessors to be more idiomatic.


sappenin:
In Rust, it seems like the more idiomatic pattern for loading data from a struct is to use properties instead of method-accessors. In addition, since our "properties" are actually loading data from an underlying host function, it's likely that we would want to cache this data in the VM after the first call (this won't add any additional memory overhead, and will save a host-function call on subsequent property accesses in any WASM code).

To pull this off (i.e., accessing a property, loading it from the host, then returning the cached value on subsequent calls) we could use a `Cell` or a `OnceCell`, like this:

```rust
use core::cell::Cell;

struct EscrowObject {
    slot_num: i32,
    account: Cell<Option<AccountID>>,
}

impl EscrowObject {
    fn account(&self) -> Result<AccountID> {
        if let Some(account) = self.account.get() {
            return Ok(account);
        }
        
        let account = current_ledger_object::get_field(sfield::Account)?;
        self.account.set(Some(account));
        Ok(account)
    }
}
```

However, we want to ensure we're using traits like `LedgerObjectCommonFields` or `TransactionCommonFields` so that eventually we can have sugar-structs (e.g., `AccountRoot`) that implement the correct trait. So, we could do something like this:

```rust
// Base trait for anything with common fields
pub trait CommonFields {
    fn flags(&self) -> Result<u32>;
    fn ledger_entry_type(&self) -> Result<u16>;
}

// Ledger objects extend common fields
pub trait LedgerObjectCommonFields: CommonFields {
    fn get_slot_num(&self) -> i32;
    
    fn flags(&self) -> Result<u32> {
        ledger_object::get_field(self.get_slot_num(), sfield::Flags)
    }
}

// Transaction fields also extend common fields  
pub trait TransactionCommonFields: CommonFields {
    fn flags(&self) -> Result<u32> {
        current_tx::get_field(sfield::Flags)
    }
}

// Specific object types can implement multiple traits
pub trait EscrowFields: LedgerObjectCommonFields {
    fn account(&self) -> Result<AccountID> {
        ledger_object::get_field(self.get_slot_num(), sfield::Account)
    }
}
```

Note that for consistency, we'll want to align our Transaction and LedgerObject traits to use property-like naming (as the above example does).

PR #243, opened 2026-07-27
feat: Create script for ledger object code generation

### LINKED ISSUE #254: Add a typed wrapper for Oracle ledger objects
The stdlib has typed ledger-object wrappers for exactly two entry types — `AccountRoot`
(`xrpl-common-stdlib/src/objects/account_root.rs`, with the `AccountFields` accessor trait) and
`Escrow` (`xrpl-escrow-stdlib/src/ledger_objects/escrow.rs`, with `EscrowFields`). Oracle entries have
none, so reading one means either raw `unsafe` host calls or the untyped slot handle.

## What to add

An `Oracle` wrapper following the existing pattern:

- `pub struct Oracle { pub slot_num: i32 }` implementing `LedgerObjectCommonFields`, which supplies
  `get_flags` / `get_ledger_entry_type` / `path()` from one `get_slot_num` method.
- An `OracleFields: LedgerObjectCommonFields` trait with named accessors for the entry's fields —
  `Owner`, `Provider`, `AssetClass`, `LastUpdateTime`, `URI`, and navigation into
  `PriceDataSeries[i]` (`BaseAsset`, `QuoteAsset`, `AssetPrice`, `Scale`).

Mirror `account_root.rs` for structure and `objects/traits.rs` for accessor style.

## Why this is a convenience, not an unblock

The ledger-object locator work adds `LedgerObject::new(slot)` — an untyped handle that implements
`LedgerObjectCommonFields`, so nested reads already work for any entry type without a wrapper. The
oracle example uses it today:

```rust
let price = LedgerObject::new(slot)
    .path()
    .field(sfield::PriceDataSeries)
    .index(0)
    .field(sfield::AssetPrice)
    .get::<u64>();
```

A typed wrapper would make that `oracle.get_asset_price(0)` and assert at the type level that the slot
holds an Oracle rather than deferring to a runtime host error. Adding wrappers is purely additive —
code written against `LedgerObject` keeps working.

## Open question first

There are ~30 XRPL ledger entry types and wrappers for two of them, so "add one per entry type" is a
policy decision that has not been made. That belongs to #82 (Determine which abstractions this library
should contain) — worth settling there before this lands, so Oracle does not become a one-off that
implies a wrapper-per-type sweep nobody signed up for.

Related: #82, #112

ricky122-5:

Adds script that automatically generates fields, getters, tests, and definitions for ledger objects in the common lib.

### Type of Change

- [ ] Documentation update
- [ ] Release


tools/generateLedgerObjects.js
```diff
@@ -0,0 +1,1021 @@
...
+
+function snakeCase(className) {
+  return toGetterName(className).slice("get_".length)
+}
+
+////////////////////////////////////////////////////////////////////////
+//  Main
+////////////////////////////////////////////////////////////////////////
+
+async function main() {
```
ricky122-5, line 769: recommend starting here in review

scripts/check-ledger-objects-generated.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
mvadari: Having one script for checking the generated objects and a separate script for running it is bloat-y

tools/generateLedgerObjects.js
```diff
@@ -0,0 +1,1021 @@
...
+// Handles both LEDGER_ENTRY(...) and LEDGER_ENTRY_DUPLICATE(...) -- the
+// latter has the exact same argument shape and only exists to dodge a JSS
+// name collision between a transaction and a ledger entry (e.g.
+// DepositPreauth), so it's parsed identically.
+function parseLedgerEntries(macroFile) {
+  const stripped = stripComments(macroFile)
+  const entries = []
+
+  const entryRe =
+    /LEDGER_ENTRY(?:_DUPLICATE)?\(\s*(lt[A-Z0-9_]+)\s*,\s*(0x[0-9a-fA-F]+)\s*,\s*([A-Za-z0-9]+)\s*,\s*([a-z0-9_]+)\s*,\s*\(\{([\s\S]*?)\}\s*\)\)/g
```
mvadari, line 52: Add a comment with an example matching output

tools/generateLedgerObjects.js
```diff
@@ -0,0 +1,1021 @@
...
+  if (!headingMatch) return map
+
+  const start = headingMatch.index + headingMatch[0].length
+  const nextHeadingMatch = markdown.slice(start).match(/^#{2,3}\s+/m)
+  const section = nextHeadingMatch
+    ? markdown.slice(start, start + nextHeadingMatch.index)
+    : markdown.slice(start)
+
+  const rowRe =
+    /^\|\s*`?([A-Za-z0-9]+)`?[^|]*\|[^|]*\|[^|]*\|[^|]*\|(.*)\|\s*$/gm
```
mvadari, line 119: Same here

tools/generateLedgerObjects.js
```diff
@@ -0,0 +1,1021 @@
...
+  for (const [, field, description] of section.matchAll(rowRe)) {
+    map.set(field, cleanDescription(description))
+  }
+  return map
+}
+
+// Fetches (and caches) the field-description map for one entry's
+// xrpl-dev-portal page. Returns an empty map if the page or its Fields table
+// can't be found -- callers fall back to a generic doc comment in that case.
+const docPageCache = new Map()
```
mvadari: Why is a cache needed here?  It'll be a different page for each object

tools/generateLedgerObjects.js
```diff
@@ -0,0 +1,1021 @@
...
+// emitted as per-entry getters.
+// - LedgerIndex: not a real field on the object -- it's a synthetic value
+//   derived from the object's keylet/index. See the NOTE in traits.rs.
+// - Flags / LedgerEntryType: covered by LedgerObjectCommonFields /
+//   CurrentLedgerObjectCommonFields, so per-entry traits must not redeclare them.
+const EXCLUDED_FIELDS = new Set(["LedgerIndex", "Flags", "LedgerEntryType"])
+
+// The Escrow entry is excluded entirely -- its field accessors, including the
+// mutable ContractData `Data` field, are hand-written in xrpl-escrow-stdlib.
+const EXCLUDED_ENTRIES = new Set(["ltESCROW"])
```
mvadari: This doesn't make sense, you won't be able to get information about _other_ escrows without it here

tools/generateLedgerObjects.js
```diff
@@ -0,0 +1,1021 @@
...
+
+// Maps a generated entry's className to the inherent `load()` constructor it
+// should get, built on top of the matching function in
+// xrpl-common-stdlib/src/keylets.rs. `params` names the Rust fn arguments
+// (used both for the `load()` signature and for forwarding into the keylet
+// call); `keyletFn` is the keylets.rs function name. Entries with no keylet
+// function (singletons: Amendments, LedgerHashes, FeeSettings, NegativeUNL,
+// DirectoryNode; plus any entry not listed here) get no `load()` -- just
+// `new(slot_num)`.
+const KEYLET_ROUTING = {
```
mvadari: This is going to be difficult to maintain as-is (case in point: you're already missing some objects here). I'd suggest not auto-generating this part and just leaving a placeholder.

tools/generateLedgerObjects.js
```diff
@@ -0,0 +1,1021 @@
...
+
+////////////////////////////////////////////////////////////////////////
+//  Rust emission
+////////////////////////////////////////////////////////////////////////
+
+const BANNER =
+  "// GENERATED -- do not hand-edit. Run scripts/generate-ledger-objects.sh to regenerate.\n"
+
+// Blob type aliases that live in crate::types::blob.
+const BLOB_TYPES = [
```
mvadari: This should be unified with the list in `generateSFields.js`

tools/generateLedgerObjects.js
```diff
@@ -0,0 +1,1021 @@
...
+  const outputDir = process.argv[3] || path.join(__dirname, "..")
+
+  const [ledgerEntriesMacro, sfieldsMacro] = await Promise.all([
+    read(rippledSource, "include/xrpl/protocol/detail/ledger_entries.macro"),
+    read(rippledSource, "include/xrpl/protocol/detail/sfields.macro"),
+  ])
+
+  const allEntries = parseLedgerEntries(ledgerEntriesMacro)
+  const entries = allEntries.filter((e) => !EXCLUDED_ENTRIES.has(e.entryName))
+  const sfieldTypes = parseSfields(sfieldsMacro)
```
mvadari, line 779: If you need to parse SFields here anyways, I wonder if it'd be worth combining this script and the SField script. But common helper functions might be easier for now.

docs/comprehensive-guide.md
```diff
@@ -104,7 +104,8 @@ Let's create a simple escrow that releases funds when an account balance exceeds
 
 use xrpl_escrow_stdlib::current_tx::escrow_finish::EscrowFinish;
 use xrpl_common_stdlib::current_tx::traits::TransactionCommonFields;
-use xrpl_common_stdlib::objects::account_root::get_account_balance;
+use xrpl_common_stdlib::objects::AccountRoot;
+use xrpl_common_stdlib::objects::traits::AccountRootFields;
```
mvadari: Why is `AccountRootFields` being imported from `traits`?

xrpl-common-stdlib/src/objects/traits.rs
```diff
@@ -725,127 +588,122 @@ mod tests {
...
+            expect_ledger_field(&mut mock, 1, sfield::Balance, 48, 1);
 
             let _guard = setup_mock(mock);
 
-            let account = AccountRoot { slot_num: 1 };
+            let account = AccountRoot::new(1);
 
             // All mandatory fields should return Ok
             assert!(account.get_account().is_ok());
-            assert!(account.owner_count().is_ok());
```
sappenin, line 778: Note that per https://github.com/ripple/xrpl-wasm-stdlib/issues/116, we eventually want to get rid of most of the `get...` prefixes in our traits because that's not idiomatic Rust in the general case. 

I don't think we should implement that change here, but for cases that don't include the `get...` prefix, can we leave those alone?

(If that breaks the generator, or makes things more complicated, then maybe let's talk)

sappenin (changes requested):

xrpl-common-stdlib/src/objects/traits.rs
ricky122-5: I don't think anything that didn't have get regressed

mvadari: @sappenin are you suggesting removing the `get_` prefix from the getters added by the script?

sappenin: > @sappenin are you suggesting removing the get_ prefix from the getters added by the script?

Hmmm... not exactly. I just noticed we're _adding_ more `get...` prefixes in this PR, and that actually works against where we want the library to go via #116.

mvadari: Yeah, the script adds the `get_` prefix

xrpl-common-stdlib/src/objects/generated/oracle.rs
```diff
@@ -0,0 +1,175 @@
...
+use crate::objects::traits::CurrentLedgerObjectCommonFields;
+use crate::objects::traits::LedgerObjectCommonFields;
+use crate::objects::{current_ledger_object, ledger_object};
+use crate::sfield;
+use crate::types::account_id::AccountID;
+use crate::types::blob::{StandardBlob, UriBlob};
+use crate::types::uint::Hash256;
+
+/// Trait providing access to fields specific to Oracle objects in any ledger.
+pub trait OracleFields: LedgerObjectCommonFields {
```
sappenin, line 13: This file probably resolves #254 right?

sappenin (approved):

xrpl-common-stdlib/src/objects/generated/oracle.rs
ricky122-5: Oop yep, didn't know it had this

xrpl-common-stdlib/src/objects/test_utils.rs
```diff
@@ -0,0 +1,61 @@
+//! Shared mock-host-binding helpers used by the `#[cfg(test)]` blocks that
```
mvadari, line 1: nit: call this file `test_utils` or something

scripts/generate-ledger-objects.sh
```diff
@@ -0,0 +1,106 @@
+#!/bin/bash
+# Generate ledger-object field accessor traits from rippled source.
+#
+# This regenerates xrpl-common-stdlib/src/objects/generated/ (one file per ledger
```
mvadari, line 4: Why not put it in the `objects` folder instead of a separate `generated` folder?

xrpl-common-stdlib/src/objects/generated/account_root.rs
```diff
@@ -0,0 +1,345 @@
...
+    fn account(&self) -> Result<AccountID> {
+        ledger_object::get_field(self.get_slot_num(), sfield::Account)
+    }
+
+    /// The sequence number of the next valid transaction for this account.
+    fn sequence(&self) -> Result<u32> {
+        ledger_object::get_field(self.get_slot_num(), sfield::Sequence)
+    }
+
+    /// The account's current [XRP balance in drops][XRP, in drops], represented as a string.
```
mvadari: Script needs some cleanup here

xrpl-common-stdlib/src/objects/generated/account_root.rs
```diff
@@ -0,0 +1,345 @@
...
+    fn previous_txn_id(&self) -> Result<Hash256> {
+        ledger_object::get_field(self.get_slot_num(), sfield::PreviousTxnID)
+    }
+
+    /// The [index of the ledger][Ledger Index] that contains the transaction that most recently modified this object.
+    fn previous_txn_lgr_seq(&self) -> Result<u32> {
+        ledger_object::get_field(self.get_slot_num(), sfield::PreviousTxnLgrSeq)
+    }
+
+    /// The identifying hash of the transaction most recently sent by this account. This field must be enabled to use the `AccountTxnID` transaction field. To enable it, send an AccountSet transaction with the `asfAccountTxnID` flag enabled.
```
mvadari: Does `cargo fmt` not catch these long comment strings?

xrpl-common-stdlib/src/objects/generated/account_root.rs
```diff
@@ -0,0 +1,345 @@
...
+    fn previous_txn_lgr_seq(&self) -> Result<u32> {
+        current_ledger_object::get_field(sfield::PreviousTxnLgrSeq)
+    }
+
+    /// The identifying hash of the transaction most recently sent by this account. This field must be enabled to use the `AccountTxnID` transaction field. To enable it, send an AccountSet transaction with the `asfAccountTxnID` flag enabled.
+    fn account_txn_id(&self) -> Result<Option<Hash256>> {
+        current_ledger_object::get_field_optional(sfield::AccountTxnID)
+    }
+
+    /// The address of a key pair that can be used to sign transactions for this account instead of the master key. Use a [SetRegularKey transaction][] to change this value.
```
mvadari: Ditto here

scripts/generate-ledger-objects.sh
```diff
@@ -0,0 +1,106 @@
...
+CHECK=0
+if [[ "${1:-}" == "--check" ]]; then
+    CHECK=1
+    shift
+fi
+
+# Default rippled source (can be overridden with the first argument, or the
+# RIPPLED_BRANCH env var for just swapping the branch/tag/commit of the
+# default XRPLF/rippled repo).
+DEFAULT_SOURCE="https://github.com/XRPLF/rippled/tree/${RIPPLED_BRANCH:-xrplf/smart-contracts}"
```
mvadari:
```suggestion
DEFAULT_SOURCE="https://github.com/XRPLF/rippled/tree/${RIPPLED_BRANCH:-ripple/se/supported}"
```

xrpl-common-stdlib/src/objects/generated/account_root.rs
ricky122-5: Added script logic to clean that up

xrpl-common-stdlib/src/objects/generated/account_root.rs
ricky122-5: Nope :( . Added script logic to wrap it

tools/generateLedgerObjects.js
```diff
@@ -0,0 +1,818 @@
...
+  //     {sfDestination, soeREQUIRED},
+  //   }))
+  const entryRe =
+    /LEDGER_ENTRY(?:_DUPLICATE)?\(\s*(lt[A-Z0-9_]+)\s*,\s*(0x[0-9a-fA-F]+)\s*,\s*([A-Za-z0-9]+)\s*,\s*([a-z0-9_]+)\s*,\s*\(\{([\s\S]*?)\}\s*\)\)/g
+
+  for (const match of stripped.matchAll(entryRe)) {
+    const [, entryName, classId, className, rpcName, fieldsBlock] = match
+    const fields = []
+    // The SoeRequired/SoeOptional/SoeDefault token appears in two rippled
+    // dialects -- `soeREQUIRED` (e.g. xrplf/smart-contracts) and `SoeRequired`
```
mvadari: Just match the latter, we should update the `xrplf/smart-contracts` branch anyways

xrpl-common-stdlib/src/objects/generated/amm.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    /// The index of the ledger that contains the transaction that most recently modified this
+    /// entry.
+    fn previous_txn_lgr_seq(&self) -> Result<Option<u32>> {
+        ledger_object::get_field_optional(self.get_slot_num(), sfield::PreviousTxnLgrSeq)
+    }
+}
+
+/// Trait providing access to fields specific to the current AMM object.
+pub trait CurrentAMMFields: CurrentLedgerObjectCommonFields {
```
mvadari, line 64: We renamed `current` to `home` in the host functions (this can be done in a separate PR if there's a bunch of upstream changes that need to be updated)

xrpl-common-stdlib/src/objects/generated/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+pub mod pay_channel;
+pub mod permissioned_domain;
+pub mod ripple_state;
+pub mod signer_list;
+pub mod ticket;
+pub mod vault;
+pub mod xchain_owned_claim_id;
+pub mod xchain_owned_create_account_claim_id;
+
+pub use account_root::{AccountRoot, AccountRootFields, CurrentAccountRootFields};
```
mvadari: Why do we have _both_ of these exports?

xrpl-common-stdlib/src/objects/mod.rs
```diff
@@ -41,3 +43,26 @@ pub use crate::fields::current_ledger_obj as current_ledger_object;
 pub use crate::fields::ledger_obj as ledger_object;
 /// Untyped handle to a slot-cached ledger object, for object types with no typed wrapper.
 pub use any_object::LedgerObject;
+
+pub use generated::{
```
mvadari, line 49: Same here

xrpl-common-stdlib/src/objects/test_utils.rs
```diff
@@ -0,0 +1,61 @@
...
+/// other field type ignores the lead byte's value here (they only care about the length), so
+/// `0x80` is a safe uniform fill.
+fn write_present(out_buff_ptr: *mut u8, out_buff_len: usize) {
+    unsafe {
+        core::ptr::write_bytes(out_buff_ptr, 0, out_buff_len);
+        if out_buff_len > 0 {
+            *out_buff_ptr = 0x80;
+        }
+    }
+}
```
mvadari, line 28: This can be a follow-up PR, but we should really be providing slightly more legitimate dummy data - i.e. there should be a certain set of bytes based on the SType

xrpl-common-stdlib/src/objects/generated/escrow.rs
```diff
@@ -0,0 +1,180 @@
...
+    fn transfer_rate(&self) -> Result<Option<u32>> {
+        ledger_object::get_field_optional(self.get_slot_num(), sfield::TransferRate)
+    }
+
+    /// The ledger index of the issuer's directory node associated with the `Escrow`. Used when the
+    /// issuer is neither the source nor destination account.
+    fn issuer_node(&self) -> Result<Option<u64>> {
+        ledger_object::get_field_optional(self.get_slot_num(), sfield::IssuerNode)
+    }
+}
```
mvadari, line 114: Why is `Data` not here?

CLAUDE.md
```diff
@@ -120,6 +120,12 @@ src/
...
...pl-common-stdlib/src/objects/generated/` holds one file per XRPL ledger-entry type (`oracle.rs`, `account_root.rs`, `bridge.rs`, ...) plus `mod.rs` (module declarations, flat re-exports, and a `//!` header listing every field whose XRPL wire type has no typed Rust mapping yet, grouped by wire type). All of it — including a `#[cfg(test)] mod tests` block per entry — is produced by `tools/generateLedgerObjects.js` from rippled's `ledger_entries.macro`/`sfields.macro`, invoked via `./scripts/generate-ledger-objects.sh`. Escrow is generated **slot-only** (its `EscrowFields` trait + `Escrow` struct are emitted like any other entry, so contracts can read _other_ escrows), but its current-object accessors and the host-mutable `Data` (`ContractData`) field stay hand-written in `xrpl-escrow-stdlib` (`CurrentEscrowFields`, and the `EscrowContractData` extension trait for `Data`). No `load()` constructor is generated (keylet inputs vary per entry — construct with `new(slot)` after caching a keylet from `keylets.rs`). Do not hand-edit anything under `generated/` — `./scripts/generate-ledger-objects.sh --check` regenerates and diffs it in CI; fix drift by changing the generator, not the output.
```
mvadari: To make this easier to maintain, the `Escrow` line should be more general to any object

xrpl-common-stdlib/src/objects/generated/amm.rs
ricky122-5: I think this should be done in another PR so that everything "Current" can be updated at once and so there's no scope creep

xrpl-common-stdlib/src/objects/generated/mod.rs
ricky122-5: Fixing, thanks!

xrpl-common-stdlib/src/objects/test_utils.rs
ricky122-5: Agreed, this 0x80 is just placeholder. We can make some robust dummy data

xrpl-common-stdlib/src/objects/generated/escrow.rs
ricky122-5: Data for escrows uses ContractData, which doesn't decode using FieldDecoder + FromLedger. The getter generation here needs that, which would make a generated getter essentially `Option<StandardBlob>` like what DID does. Instead, we add an extension trait `EscrowContractData` that has the read/write functionality

ricky122-5: aliased ContractData as Blob<1024>

mvadari (approved):

PR #244, opened 2026-07-28
fix: Update contract data size to 1024

ricky122-5:

Updates the contract data size to 1024, following the spec

### Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Release

### DIFF
--- a/xrpl-common-stdlib/src/types/mod.rs
+++ b/xrpl-common-stdlib/src/types/mod.rs
@@ -14,9 +14,3 @@ pub mod opaque_float;
 pub mod public_key;
 pub mod transaction_type;
 pub mod uint;
-
-// TODO: Move these to the `types` crate.
-// Relocated from the old top-level `src/types.rs`, which collided with this
-// directory's module path once `core::types` was promoted to `crate::types`.
-pub const XRPL_CONTRACT_DATA_SIZE: usize = 1024;
-pub type ContractData = [u8; XRPL_CONTRACT_DATA_SIZE];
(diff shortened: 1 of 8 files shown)


xrpl-common-stdlib/src/types/mod.rs
```diff
@@ -18,5 +18,5 @@ pub mod uint;
 // TODO: Move these to the `types` crate.
 // Relocated from the old top-level `src/types.rs`, which collided with this
 // directory's module path once `core::types` was promoted to `crate::types`.
-pub const XRPL_CONTRACT_DATA_SIZE: usize = 4096; //TODO size??
+pub const XRPL_CONTRACT_DATA_SIZE: usize = 1024;
```
TimothyBanks: I see this definition in contract_data.rs as well (in the main branch).  Are these different definitions?

sappenin: I think they're the same. The style in this project is to locate constants close to their struct definition, so let's just remove this one and use the one in `contract_data.rs` (unless that file is slated to go away for some reason?)

sappenin (changes requested):

sappenin (approved):

xrpl-common-stdlib/src/types/mod.rs
sappenin: @TimothyBanks I think this is fixed now with the latest commits. If so, please resolve this one.

TimothyBanks (approved):

PR #247, opened 2026-07-28
fix: Stop copying bytes in FieldDecoder

sappenin:
FieldDecoder::decode now takes Self::Buffer by value instead of &Self::Buffer, so implementations can move the host-filled buffer directly into the returned type instead of copying it — no more copying of bytes that don't need to be copied.

Every caller (fields::current_tx::get_field[_optional], fields::ledger_obj::get_field[_optional]) already owns its buf locally and never touches it again after handing it to decode_result/decode, so passing it by value is free — it just makes that existing ownership explicit at the type level instead of forcing a copy through a borrow.


ricky122-5 (approved):
LGTM

ricky122-5 (approved):

PR #249, opened 2026-07-29
fix: Reduce Blob byte copies

sappenin:
Adds Blob<N>-specific zero-copy accessors — get_blob_field/get_blob_field_optional in both fields::current_tx and fields::ledger_obj — that write host bytes directly into the returned Blob<N>'s own memory, eliminating an avoidable memcpy on every large-blob field read.

The generic get_field/get_field_optional (used by every FieldDecoder type: u8..u64, AccountID, Amount, UInt<N>, TransactionType, and Blob<N>) allocate their own local buffer, have the host write into it, then reconstruct the returned T from that buffer. For small fixed-size types that reconstruction is reliably optimized away. For Blob<N> it is not — Blob<N> is { data: [u8; N], len: usize }, a different memory shape than the bare [u8; N] the generic path allocates, so building the Blob is a genuine field-by-field reconstruction rather than a reinterpret.

Confirmed empirically compiling the actual generic-path retrieval of a WasmBlob (N = 4096) against this crate's real release profile (opt-level = "s", lto = true, codegen-units = 1) and disassembling with wasm2wat shows a genuine ~4092-byte memcpy per read — -Os doesn't fuse the redundant copy away the way -O2/-O3 might.

### DIFF
--- a/xrpl-common-stdlib/src/fields/ledger_obj.rs
+++ b/xrpl-common-stdlib/src/fields/ledger_obj.rs
@@ -7,8 +7,10 @@
 
 use crate::fields::decoder::{FromLedger, decode_result};
 use crate::host::error_codes::FIELD_NOT_FOUND;
-use crate::host::{Result, get_ledger_obj_field};
+use crate::host::{Error, Result, get_ledger_obj_field};
 use crate::sfield::SField;
+use crate::types::blob::Blob;
+use core::mem::MaybeUninit;
 
 /// Retrieves a field from the ledger object cached in `slot` using an SField constant.
 ///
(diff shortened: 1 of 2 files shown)


xrpl-common-stdlib/src/fields/ledger_obj.rs
```diff
@@ -51,9 +53,86 @@ pub fn get_field_optional<T: FromLedger, const CODE: i32>(
     decode_result::<T>(buf, n).map(Some)
 }
 
+// --- `Blob<N>`-specific accessors ------------------------------------------------------------
```
ricky122-5, line 51: If you want to keep these comments that's fine, just pointing them out as kind of a big block

ricky122-5 (approved):
LGTM

ricky122-5 (approved):

PR #251, opened 2026-07-29
feat: Ledger obj field decoder migration

ricky122-5:

Migrates LedgerObj - accessible fields to FieldDecoder and gets rid of LedgerObjGetter

### Type of Change

- [ ] Documentation update
- [ ] Release


xrpl-common-stdlib/src/fields/current_ledger_obj.rs
```diff
@@ -0,0 +1,322 @@
...
+
+/// Retrieves a field from the current ledger object using an SField constant.
+///
+/// # Returns
+///
+/// Returns a `Result<T>` where:
+/// * `Ok(T)` - The field value for the specified field
+/// * `Err(Error)` - If the field cannot be retrieved, has unexpected size, or fails to decode
+#[inline]
+pub fn get_field<T: FromLedger, const CODE: i32>(_: SField<T, CODE>) -> Result<T> {
```
sappenin, line 28: I'm supportive of moving these functions out of `field_helpers.rs`. However, I'm not sure if the `fields` module is the right location for this file (i.e., `current_ledger_obj.rs`). Primary reason is this isn't a "field" concern -- it's instead the "Current Ledger Object", which is more than just fieds.

What do you think about locating this file inside of `objects` module?

xrpl-common-stdlib/src/objects/mod.rs
```diff
@@ -1,1142 +1,40 @@
...
-
-#[cfg(test)]
-mod tests {
-    use super::current_ledger_object;
-    use super::ledger_object;
-    use crate::sfield;
-
-    #[test]
-    #[should_panic]
-    fn test_array_get_field_panics() {
```
sappenin, line 1097: There are a lot of test in here that are removed, which is correct. However, the number of tests added to the FieldLocator seems.... less? How is coverage looking with this changeset?

xrpl-common-stdlib/src/types/uint.rs
```diff
@@ -195,11 +77,14 @@ impl<const N: usize> FieldDecoder for UInt<N> {
...
-// already have the generic `FieldDecoder` impl above and gain `FromCurrentTx`/`FromLedger`
-// if and when a field typed to them needs to be read.
+// Readability markers are added per width only where a field of that type is actually read:
+// `Hash256` for the many 32-byte hash fields (transaction IDs, ledger indexes, ...) and `Hash128`
+// for the 16-byte `EmailHash` ledger field. All widths already share the generic `FieldDecoder`
+// impl above; the remaining widths (`Hash160`/`Hash192`) gain markers if and when a field typed
+// to them needs to be read.
 impl FromCurrentTx for Hash256 {}
 impl FromLedger for Hash256 {}
+impl FromLedger for Hash128 {}
```
sappenin, line 84: Should we add this for the other HashTypes?

```suggestion
impl FromLedger for Hash192 {}
impl FromLedger for Hash160 {}
impl FromLedger for Hash128 {}
```

xrpl-common-stdlib/src/types/uint.rs
```diff
@@ -195,11 +77,14 @@ impl<const N: usize> FieldDecoder for UInt<N> {
...
 
-// Only `Hash256` is marked readable; the narrower hash widths (`Hash128`/`Hash160`/`Hash192`)
-// already have the generic `FieldDecoder` impl above and gain `FromCurrentTx`/`FromLedger`
-// if and when a field typed to them needs to be read.
+// Readability markers are added per width only where a field of that type is actually read:
+// `Hash256` for the many 32-byte hash fields (transaction IDs, ledger indexes, ...) and `Hash128`
+// for the 16-byte `EmailHash` ledger field. All widths already share the generic `FieldDecoder`
+// impl above; the remaining widths (`Hash160`/`Hash192`) gain markers if and when a field typed
+// to them needs to be read.
 impl FromCurrentTx for Hash256 {}
```
sappenin, line 83: Should we add this for the other HashTypes?

sappenin (changes requested):

xrpl-common-stdlib/src/fields/current_ledger_obj.rs
ricky122-5: Objects/mod.rs reexports it already. I'd say it should stay in fields bc it's sharing decode machinery with decoder.rs/locator.rs/current_tx.rs

It's the field logic in context of ledger objects, like how current_tx is the field retrieval logic for current transactions
Still open to discussion though, wdyt

xrpl-common-stdlib/src/objects/mod.rs
ricky122-5: We were able to reduce scope on tests by testing the shared FieldDecoder path rather than the LedgerObjGetter per-type, so equivalent coverage and different shape.

xrpl-common-stdlib/src/types/uint.rs
ricky122-5: Yep, will add

xrpl-common-stdlib/src/fields/current_ledger_obj.rs
sappenin: Ok, I see what you're trying to do (I misunderstood the `fields` module a bit). Let's leave it for now and I'll bring it up later if I think it needs to change.

sappenin (approved):

PR #217, opened 2026-07-10
feat: Locator builder for ledger obj, shared pathbuf

ricky122-5:
## High Level Overview of Change

Expands the fluent Locator builder to ledger objects as well. Tx and Ledger objects share a PathBuf struct that encapsulates and matches contexts to appropriate host functions.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactor (non-breaking change that only restructures code)
- [x] Tests (added tests for existing code, or tests for the new feature in this PR)
- [x] Documentation update
- [ ] Build / CI / tooling change
- [ ] Release

PR #253, opened 2026-07-30
feat: Ledger locator path + shared path builder

ricky122-5:

Adds the fluent `Locator` path builder for ledger objects (`LedgerPathBuilder`), reachable from any
ledger object via `obj.path()` / `ctx.escrow().path()`.

Adds locator path for ledger objects as well as a shared path builder

### Context of Change

**The shared core is scoped to the ledger side, unlike #217.** #217 extracted a private `PathBuf`
with a 3-way `NestedSource { Tx, CurrentLedgerObj, LedgerObj }` enum and rewrote `TxPathBuilder` as a
newtype over it. Ported forward as-is, that has two costs the tx side shouldn't pay:

- `TxPathBuilder` grows 80 → 88 bytes (measured), and it is moved by value on every `.field()` /
  `.index()`, up to 16 per path.
- `PathBuf::get::<T>` becomes one monomorphization shared by all three sources, so selecting the host
  function turns into a runtime `match` on a discriminant that is statically known at construction.
  It very likely folds under `lto = true, codegen-units = 1`, but #252 emitted a direct call and this
  would make it depend on the optimizer — on a gas-metered target.

More importantly it weakens an invariant. In #252, "a tx builder reads tx fields" was
*unrepresentable otherwise* — there was one `read`, hard-wired. With a 3-way enum,
`LedgerPathBuilder(PathBuf::new(NestedSource::Tx))` type-checks, and the guarantee degrades to a
convention held by two constructors. The trybuild snapshots don't catch that; they pin the public
bound, which is the part that didn't regress.

So the enum here holds only the two *ledger* sources:

```rust
enum LedgerSource { Current, Slot(i32) }   // no Tx variant exists to select
```
TxPathBuilder is left byte-for-byte as #252 shipped it — 80 bytes, direct get_tx_nested_field,
zero diff to its struct or impl. "A LedgerPathBuilder cannot read a transaction field" is now true
by construction. The two ledger sources genuinely need a runtime discriminant: they share the
FromLedger bound so they must be one public type, and Slot carries runtime state. Two sources ×
two terminals (get and array_len) is what earns the factored-out LedgerPath. Cost is ~35
duplicated lines between the builders, which is the trade I'd rather make than churn just-merged code.

I did consider a trait-parameterized Path<S: NestedSource> that dedups those lines while keeping tx
a ZST with direct dispatch. It can't recover the marker bound at the type level either — Rust has no
associated traits, so T: S::Readable isn't expressible and the bound stays on the public wrapper
in every variant of this design. It buys line count only, which isn't worth a generic + trait
indirection in a no_std stdlib. Worth revisiting if array_len() is later added to
TxPathBuilder, since two terminals × two builders pushes duplication to ~60 lines.

Why LedgerObject exists. LedgerPathBuilder::for_ledger_obj(slot) is pub(crate), so the only
public door is LedgerObjectCommonFields::path() — which needs a type implementing that trait. The
repo has exactly two (AccountRoot, Escrow). The one real consumer of slot-based nested reads is
the oracle example, reading off an Oracle entry, which has no wrapper. Without LedgerObject the
feature ships with no reachable caller and cannot replace the unsafe it exists to replace. Typed
wrappers remain the better choice when you want named accessors; adding one later is purely additive.

Why array_len and not len. Locator::len() already means "bytes packed into the path". This
returns the length of the array the path points at, so the longer name avoids a genuinely ambiguous
call. Returns Ok(0) for a present-but-empty array rather than treating that as an error.

Smaller fixes. index() now rejects values above i32::MAX instead of packing them as a negative
the host would read back as a field code. The get_optional rustdoc documents a pre-existing gap
inherited from #252: variable-length fields are reported as a zero-byte write rather than
FieldNotFound, so an absent Blob<N> yields Ok(Some(blob)) with len == 0, never Ok(None) —
the same distinction ledger_obj::get_blob_field_optional exists to handle. Fixing that needs a
blob-aware terminal applied uniformly across both builders and both crates, so it's deferred rather
than half-done here.

### Type of Change

- [ ] Documentation update
- [ ] Release

### DIFF
--- a/examples/smart-escrows/oracle/src/lib.rs
+++ b/examples/smart-escrows/oracle/src/lib.rs
@@ -3,11 +3,11 @@
 #[cfg(not(target_arch = "wasm32"))]
 extern crate std;
 
-use xrpl_common_stdlib::fields::locator::Locator;
-use xrpl_common_stdlib::host::error_codes::match_result_code;
-use xrpl_common_stdlib::host::trace::{DataRepr, trace_data, trace_num};
-use xrpl_common_stdlib::host::{Result, Result::Err, Result::Ok};
+use xrpl_common_stdlib::host::trace::{trace, trace_num};
+use xrpl_common_stdlib::host::{Error, Result, Result::Err, Result::Ok};
 use xrpl_common_stdlib::keylets::oracle_keylet;
+use xrpl_common_stdlib::objects::LedgerObject;
+use xrpl_common_stdlib::objects::traits::LedgerObjectCommonFields;
 use xrpl_common_stdlib::r_address;
 use xrpl_common_stdlib::types::account_id::AccountID;
 use xrpl_common_stdlib::{host, sfield};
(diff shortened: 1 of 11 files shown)


examples/smart-escrows/oracle/src/lib.rs
```diff
@@ -18,35 +18,45 @@ const ORACLE_OWNER: AccountID = r_address!("rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh")
 const ORACLE_DOCUMENT_ID: u32 = 1;
 
 pub fn get_price_from_oracle(slot: i32) -> Result<u64> {
-    let mut locator = Locator::new();
-    locator.pack(sfield::PriceDataSeries);
-    locator.pack(0);
-    locator.pack(sfield::AssetPrice);
+    // The Oracle entry has no typed wrapper, so reach its nested fields through the untyped slot
```
sappenin, line 21: > The Oracle entry has no typed wrapper

Is there a task to add a typed wrapper? If not, can you capture in the project's issue list?

sappenin (approved):

examples/smart-escrows/oracle/src/lib.rs
ricky122-5: not ATM. will add as an issue.

PR #255, opened 2026-07-30
refactor: Host function errors and trace functions

pwang200:
Matches the xrpld-side trace refactor. xrpld declares all five trace host functions as `void`, so the stdlib's `-> i32` signature never described the real ABI. They now return nothing, through the `extern "C"` block, the `HostBindings` trait, all three impls, and the public `trace*` helpers. Call sites drop their `let _ =`, which `clippy::let_unit_value` makes an error under `-Dwarnings`.

Also realigns two error codes xrpld reassigned: `-1` is `Unimplemented`, not an internal error, and `-11` is `OutOfTransferLimit`, not a decoding failure. Both displaced concepts are still needed here, so they move into a library-owned range growing upward from `i32::MIN` (`INTERNAL_ERROR` = `i32::MIN`, `INVALID_DECODING` = `i32::MIN + 1`) where they can't collide with a host code. Host codes are the host/guest interface; conditions the stdlib detects itself belong in its own range. `DecodeError` still maps to `InvalidDecoding`, so only the value moves — no call site renames.

`tools/compareHostFunctions.js` required a mandatory `-> T` in all four of its regexes, so unit-returning functions were skipped and then reported as missing from the Rust trait. Fixed to treat an omitted or `-> ()` return as `void`.

Pairs with the xrpld trace2 change and should land alongside it. Until then the host function audit reports five trace return-type differences, which is the expected temporary skew.

Verified: build, clippy, fmt, wasm-exports and markdown checks pass; `cargo test --workspace` is 490 passed / 0 failed. Integration tests need an xrpld node and were not run locally.

### DIFF
--- a/xrpl-common-stdlib/src/host/error_codes.rs
+++ b/xrpl-common-stdlib/src/host/error_codes.rs
@@ -2,8 +2,8 @@ use crate::host::Error::PointerOutOfBounds;
 use crate::host::trace::trace_num;
 use crate::host::{Error, Result, Result::Err, Result::Ok};
 
-/// Reserved for internal invariant trips, generally unrelated to inputs.
-pub const INTERNAL_ERROR: i32 = -1;
+/// The host function has an empty/stub implementation (not yet implemented by the host).
+pub const UNIMPLEMENTED: i32 = -1;
 /// The requested serialized field could not be found in the specified object.
 pub const FIELD_NOT_FOUND: i32 = -2;
 /// The provided buffer is too small to hold the requested data.
@@ -22,8 +22,8 @@ pub const SLOTS_FULL: i32 = -8;
 pub const EMPTY_SLOT: i32 = -9;
 /// The requested ledger object could not be found.
 pub const LEDGER_OBJ_NOT_FOUND: i32 = -10;
-/// An error occurred while decoding serialized data.
-pub const INVALID_DECODING: i32 = -11;
+/// The operation would exceed the allowed transfer limit.
+pub const OUT_OF_TRANSFER_LIMIT: i32 = -11;
 /// The data field is too large to be processed.
 pub const DATA_FIELD_TOO_LARGE: i32 = -12;
 /// A pointer or buffer length provided as a parameter described memory outside the allowed memory region.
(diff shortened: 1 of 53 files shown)


pwang200:
Failed tests are as expected, not actionable. The audits compare against old code which doesn't have the trace refactor yet.

xrpl-common-stdlib/src/host/error_codes.rs
```diff
@@ -43,6 +43,31 @@ pub const INVALID_FLOAT_INPUT: i32 = -19;
...
+// `OutOfTransferLimit`). Both concepts are still needed here, so they moved down here
+// rather than being dropped.
+// ###################################################################################
+
+/// Reserved for internal invariant trips in the stdlib, generally unrelated to inputs.
+/// These indicate a bug and should be reported with an issue. Note that the invariant trips
+/// the stdlib can detect eagerly panic instead (see
+/// [`match_result_code_with_expected_bytes`]), so this is currently a reserved slot rather
+/// than one any library code returns.
+pub const INTERNAL_ERROR: i32 = i32::MIN;
```
sappenin: Remind me why we need to reserve these (either now, or in the future)? E.g., seeing this error doesn't really tell me anything, so how is this any better than just a panic?

xrpl-common-stdlib/src/host/host_bindings_trait.rs
```diff
@@ -1545,7 +1541,7 @@ pub trait HostBindings {
         data_read_ptr: *const u8,
         data_read_len: usize,
         as_hex: i32,
-    ) -> i32;
```
sappenin, line 1534: Doesn't trace actually return the number of bytes written? Or are we (as part of this PR) just saying that doesn't happen anymore?

Also, does the work @kuznetsss is doing have any bearing on this PR?

pwang200: Removing the return value from trace_* host functions was agreed by all of the team as part of the "more tec code, finish function return value meaning, error code and trace_* functions refactor" task. 

Edit: It seems we need more discussion.

xrpl-common-stdlib/src/host/error_codes.rs
pwang200: There was a standup discussion about the scope of this PR, i.e. how much this PR should refactor the error codes. At that time, we lean towards smaller PR to match the xrpld side changes and leave error code refactoring (about errors that are not from xrpld host functions) to a separate PR.  

About `INTERNAL_ERROR`, I think it is not used in production code, but used a lot in unit tests to signal something is wrong. Claude and I both like "SOME_ERROR" error but we could not agree on how to implement it. (I want to simply replace `INTERNAL_ERROR` with it (i.e. pub const SOME_ERROR: i32 = i32::MIN;). Claude doesn't like it because it is test only, and doesn't deserve a value in production code. make sense. but I don't like its suggested implementation....) I felt this kind of discussion belongs to a separate PR about stdlib side errors.

xrpl-common-stdlib/src/host/host_bindings_trait.rs
mvadari: We switched it from number of bytes to 0 a while ago, to prevent consensus issues based on debug settings

sappenin: Resolving per our latest conversations around changes to the `trace` functions.

xrpl-common-stdlib/src/host/error_codes.rs
sappenin: What if we compromised and just remove `INVALID_DECODING`? 

(I mean, I don't think reserving `INTERNAL_ERROR` is adding any more value on top of just a panic, but if you really want that, I could tolerate it).

pwang200: I want to remove `INTERNAL_ERROR`, but don't have a good way to do it yet. It is not used in production code. It is used in unittests to signal something wrong. If we are not testing with panic, then should not replace it with panic. In other words, it is a place-holder for unittests when the authors of the tests simply wanted An error with no real meaning (I guess).  

`INVALID_DECODING` seems to have legitimate use case. 

Overall, I think the mismatching is result from that the stdlib is not a thin wrapper of the host functions (anymore). It has its own logic and can have error cases that host functions wont have. It has been using host functions errors since that was a bigger set, but not anymore. I think the right path forward is to have stdlib's own error codes. E.g. the decoding error was from Craft iirc, for good reason, that never existed in rippled side.

sappenin: > I want to remove INTERNAL_ERROR, but don't have a good way to do it yet. It is not used in production code.

Ah, I see. If that's the case, then can't the code only live in test code, or even just be a number (e.g., `i32::MIN` used in test assertions) instead of something exported or reserved by the library? If it's just used in unit tests, then there's no reason to define it here.

> the stdlib is not a thin wrapper of the host functions (anymore). It has its own logic and can have error cases that host functions wont have...I think the right path forward is to have stdlib's own error codes

I don't quite agree, but that's just because I'm not convinced there's a use-case for knowing that, for example, a Smart Escrow failed and returned a value of `i32::MIN`.

sappenin (approved):

PR #256, opened 2026-08-03
feat: Newtype wrappers

ricky122-5:

Adds newtype wrappers for messages, signatures, and accountID variations to guard against errors in adjacent arguments

### Type of Change

- [ ] Documentation update
- [ ] Release


mvadari (dismissed):
This PR requires more conversation before we make this change. I understand @kuznetsss's reasoning on preventing typos etc, but this change increases the overhead for users working with the stdlib, and that concerns me.

ricky122-5:
> This PR requires more conversation before we make this change. I understand @kuznetsss's reasoning on preventing typos etc, but this change increases the overhead for users working with the stdlib, and that concerns me.

Fair, was also concerned about that. Looked into a couple of things that sort of justified it enough to me to make a PR:

1. If the user just tries to put in a raw AccountID, rustc points them with the suggested fix, and it's all a compile error. the tradeoff is that it's passed by-value which in contrast to borrowed &AccountID is asymmetric
```
error[E0308]: mismatched types
25 |  match credential_keylet(&account_id, account_id, cred_type) {
   |                                       ^^^^^^^^^^ expected `Issuer`, found `AccountID`
help: try wrapping the expression in `...::Issuer`
25 |  match credential_keylet(&account_id, ...::Issuer(account_id), cred_type) {
   |                                       ++++++++++++++++++       +
```
2. New users are likely not going to touch these. The AccountID wrappers are used when someone computes a credential/delegate/paychan keylet, not something the average contract may do. The wrappers for check_sig may be used more, but then I feel the wrappers are net positive still

Open to discussion

TimothyBanks (approved):

ricky122-5:
Remove From trait, remove AccountID types

sappenin (approved):

TimothyBanks (approved):

PR #259, opened 2026-08-03
feat: Revamp Number, add conversion host function wrappers

ricky122-5:

Adds float host function wrappers in the Number class.

### Type of Change

- [ ] Documentation update
- [ ] Release

### DIFF
--- a/xrpl-common-stdlib/src/types/number.rs
+++ b/xrpl-common-stdlib/src/types/number.rs
@@ -5,10 +5,7 @@ use crate::host::error_codes::match_result_code_with_expected_bytes;
 /// The number of bytes in the serialized STNumber (float) representation.
 const NUMBER_SIZE: usize = 12;
 
-/// Rounding mode used for all `Number` operations.
-///
-/// The first cut of this API rounds to nearest (matching rippled's default). Directed-rounding
-/// variants can be layered on later without breaking callers.
+/// Rounding mode for `Number` conversions: to nearest, matching rippled's default.
 const TO_NEAREST: i32 = host::FLOAT_ROUNDING_MODES_TO_NEAREST;
 
 /// An opaque XRPL `STNumber` value: a decimal float represented as `mantissa × 10^exponent`.
@@ -26,7 +23,7 @@ const TO_NEAREST: i32 = host::FLOAT_ROUNDING_MODES_TO_NEAREST;
 ///
 /// `PartialEq`/`Eq` compare the raw bytes. The host canonicalizes every value it emits, so bytewise
 /// equality matches semantic equality for host-produced values.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 #[repr(C)]
 pub struct Number(pub [u8; NUMBER_SIZE]);
 
@@ -45,7 +42,7 @@ impl Number {
     ]);
 
     /// Converts a signed integer to a `Number`.
-    pub fn float_from_int(value: i64) -> Result<Number> {
+    pub fn from_int(value: i64) -> Result<Number> {
         let mut out = [0u8; NUMBER_SIZE];
         let rescode =
             unsafe { host::float_from_int(value, out.as_mut_ptr(), NUMBER_SIZE, TO_NEAREST) };
@@ -53,7 +50,7 @@ impl Number {
     }
 
     /// Converts an unsigned integer to a `Number`.
-    pub fn float_from_uint(value: u64) -> Result<Number> {
+    pub fn from_uint(value: u64) -> Result<Number> {
         // The host reads the unsigned value from a little-endian byte buffer (native order on the
         // little-endian WASM target).
         let value_bytes = value.to_le_bytes();
@@ -71,7 +68,7 @@ impl Number {
     }
 
     /// Constructs a `Number` from an explicit mantissa and exponent (`mantissa × 10^exponent`).
-    pub fn float_from_mant_exp(mantissa: i64, exponent: i32) -> Result<Number> {
+    pub fn from_mant_exp(mantissa: i64, exponent: i32) -> Result<Number> {
         let mut out = [0u8; NUMBER_SIZE];
         let rescode = unsafe {
             host::float_from_mant_exp(
@@ -86,7 +83,7 @@ impl Number {
     }
 
     /// Converts a serialized `STAmount` (e.g. from an amount field) to a `Number`.
-    pub fn float_from_stamount(bytes: &[u8]) -> Result<Number> {
+    pub fn from_stamount(bytes: &[u8]) -> Result<Number> {
         let mut out = [0u8; NUMBER_SIZE];
         let rescode = unsafe {
             host::float_from_stamount(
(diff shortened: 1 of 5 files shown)


xrpl-common-stdlib/src/types/number.rs
```diff
@@ -1,28 +1,300 @@
...
+///
+/// # Important
+///
+/// This type is intentionally opaque: arithmetic and conversions MUST go through the host, which
+/// delegates to rippled's `Number` class so results stay exactly consensus-compatible. The bytes
+/// are only ever produced by the host, so callers cannot construct an out-of-range value.
+///
+/// `PartialEq`/`Eq` compare the raw bytes. The host canonicalizes every value it emits, so bytewise
+/// equality matches semantic equality for host-produced values.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
sappenin: We might want to remove the default `Copy` because this tends to duplicate 12 bytes whenever a new assignment happens. Better to move this value, or pass a reference.

xrpl-common-stdlib/src/types/number.rs
```diff
@@ -1,28 +1,300 @@
...
+        0x0D, 0xE0, 0xB6, 0xB3, 0xA7, 0x64, 0x00, 0x00, 0xFF, 0xFF, 0xFF, 0xEE,
+    ]);
+
+    /// The value `-1` (mantissa = -1,000,000,000,000,000, exponent = -15).
+    pub const NEGATIVE_ONE: Number = Number([
+        0xF2, 0x1F, 0x49, 0x4C, 0x58, 0x9C, 0x00, 0x00, 0xFF, 0xFF, 0xFF, 0xEE,
+    ]);
+
+    /// Converts a signed integer to a `Number`.
+    pub fn float_from_int(value: i64) -> Result<Number> {
```
sappenin: The `Number` struct is not the same thing as the FFI layer. So, I suggest we make these functions "Number-specific" by removing the `float_` prefix, like this:


```suggestion
    pub fn from_int(value: i64) -> Result<Number> {
```

xrpl-common-stdlib/src/types/number.rs
```diff
@@ -1,28 +1,300 @@
...
+        let mut out = [0u8; NUMBER_SIZE];
+        let rescode =
+            unsafe { host::float_from_int(value, out.as_mut_ptr(), NUMBER_SIZE, TO_NEAREST) };
+        match_result_code_with_expected_bytes(rescode, NUMBER_SIZE, || Number(out))
+    }
+
+    /// Converts an unsigned integer to a `Number`.
+    pub fn float_from_uint(value: u64) -> Result<Number> {
```
sappenin: The `Number` struct is not the same thing as the FFI layer. So, I suggest we make these functions "Number-specific" by removing the `float_` prefix, like this:


```suggestion
    pub fn from_uint(value: u64) -> Result<Number> {
```

xrpl-common-stdlib/src/types/number.rs
```diff
@@ -1,28 +1,300 @@
...
+                out.as_mut_ptr(),
+                NUMBER_SIZE,
+                TO_NEAREST,
+            )
+        };
+        match_result_code_with_expected_bytes(rescode, NUMBER_SIZE, || Number(out))
+    }
+
+    /// Constructs a `Number` from an explicit mantissa and exponent (`mantissa × 10^exponent`).
+    pub fn float_from_mant_exp(mantissa: i64, exponent: i32) -> Result<Number> {
```
sappenin: The `Number` struct is not the same thing as the FFI layer. So, I suggest we make these functions "Number-specific" by removing the `float_` prefix, like this:


```suggestion
    pub fn from_mant_exp(mantissa: i64, exponent: i32) -> Result<Number> {
```

xrpl-common-stdlib/src/types/number.rs
```diff
@@ -1,28 +1,300 @@
...
+                out.as_mut_ptr(),
+                NUMBER_SIZE,
+                TO_NEAREST,
+            )
+        };
+        match_result_code_with_expected_bytes(rescode, NUMBER_SIZE, || Number(out))
+    }
+
+    /// Converts a serialized `STNumber` (12-byte) to a `Number`.
+    pub fn float_from_stnumber(bytes: &[u8]) -> Result<Number> {
```
sappenin: The `Number` struct is not the same thing as the FFI layer. So, I suggest we make these functions "Number-specific" by removing the `float_` prefix, like this:


```suggestion
    pub fn from_stnumber(bytes: &[u8]) -> Result<Number> {
```

xrpl-common-stdlib/src/types/number.rs
```diff
@@ -1,28 +1,300 @@
...
+                out.as_mut_ptr(),
+                NUMBER_SIZE,
+                TO_NEAREST,
+            )
+        };
+        match_result_code_with_expected_bytes(rescode, NUMBER_SIZE, || Number(out))
+    }
+
+    /// Converts this `Number` to a signed integer, rounding to nearest.
+    pub fn float_to_int(&self) -> Result<i64> {
```
sappenin: The `Number` struct is not the same thing as the FFI layer. So, I suggest we make these functions "Number-specific" by removing the `float_` prefix, like this:


```suggestion
    pub fn to_int(&self) -> Result<i64> {
```

xrpl-common-stdlib/src/types/number.rs
```diff
@@ -1,28 +1,300 @@
...
+                TO_NEAREST,
+            )
+        };
+        match_result_code_with_expected_bytes(rescode, 8, || i64::from_le_bytes(int_bytes))
+    }
+
+    /// Decomposes this `Number` into its `(mantissa, exponent)` components.
+    ///
+    /// No rounding is applied — the value is already rounded/canonical.
+    pub fn float_to_mant_exp(&self) -> Result<(i64, i32)> {
```
sappenin: The `Number` struct is not the same thing as the FFI layer. So, I suggest we make these functions "Number-specific" by removing the `float_` prefix, like this:


```suggestion
    pub fn to_mant_exp(&self) -> Result<(i64, i32)> {
```

sappenin (changes requested):

xrpl-common-stdlib/src/types/number.rs
```diff
@@ -1,28 +1,294 @@
...
-}
+use crate::host;
+use crate::host::Result;
+use crate::host::error_codes::match_result_code_with_expected_bytes;
+
+/// The number of bytes in the serialized STNumber (float) representation.
+const NUMBER_SIZE: usize = 12;
+
+/// Rounding mode for `Number` conversions: to nearest, matching rippled's default.
+const TO_NEAREST: i32 = host::FLOAT_ROUNDING_MODES_TO_NEAREST;
```
sappenin: WDYT about just using the enum for these values, to prevent duplication?

sappenin (changes requested):

xrpl-common-stdlib/src/types/number.rs
ricky122-5: Yep, will do. thanks!

sappenin (approved):

PR #260, opened 2026-08-03
feat: Float arithmetic host function wrappers

ricky122-5:

Adds float operation host function wrappers to `Number`.

### Type of Change

- [ ] Documentation update
- [ ] Release

### DIFF
--- a/xrpl-common-stdlib/src/types/number.rs
+++ b/xrpl-common-stdlib/src/types/number.rs
@@ -1,7 +1,6 @@
 use crate::host;
-use crate::host::Result;
-use crate::host::RoundingMode;
 use crate::host::error_codes::match_result_code_with_expected_bytes;
+use crate::host::{Error, Result, RoundingMode};
 
 /// The number of bytes in the serialized STNumber (float) representation.
 const NUMBER_SIZE: usize = 12;
@@ -23,7 +22,7 @@ const NUMBER_SIZE: usize = 12;
 /// equality matches semantic equality for host-produced values.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[repr(C)]
-pub struct Number(pub [u8; NUMBER_SIZE]);
+pub struct Number([u8; NUMBER_SIZE]);
 
 impl Number {
     /// The value `0`.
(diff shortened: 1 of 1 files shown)


xrpl-common-stdlib/src/types/iou_number.rs
```diff
@@ -32,8 +32,8 @@
 /// - `Debug, Clone`: Standard traits for development and consistency
 ///
 /// **Note**: `PartialEq` and `Eq` perform bitwise comparison only. For semantic comparison of
-/// amounts (e.g., handling different representations of zero), convert to `Number` and use the
-/// host `compare`.
+/// amounts (e.g., handling different representations of zero), convert to `Number` and use
+/// `Number::float_compare`.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
sappenin: I'm not sold on `Copy` - anywhere that an `IouNumber` comes through, it should probably just be passed by reference or actually moved (to save Gas). Thoughts?

xrpl-common-stdlib/src/types/number.rs
```diff
@@ -153,6 +154,124 @@ impl Number {
             )
         })
     }
+
+    /// Compares this `Number` to another, returning their [`Ordering`](core::cmp::Ordering).
+    pub fn float_compare(&self, other: &Number) -> Result<core::cmp::Ordering> {
```
sappenin: I think when a developer interacts with Number in this manner, they're not going to be thinking of the FFI layer or `floats`. However, the other functions (e.g., add `add`/`divide`) have failure and rounding modes that the user will need to be aware of, which justifies them having their own functions. 

However, in this case for comparison, we have no such failure modes and can instead use standard Rust idioms instead, like this:

```suggestion
/// Compares this `Number` to another, returning their [`Ordering`](core::cmp::Ordering).
    ///
    /// Backs the [`Ord`]/[`PartialOrd`] impls below. Not exposed publicly: unlike the arithmetic
    /// `float_*` operations, comparing two well-formed (host-canonicalized) `Number`s has no
    /// legitimate failure mode, so there's no reason to push a fallible API onto callers.
    fn compare_via_host(&self, other: &Number) -> Result<core::cmp::Ordering> {
```

Then:

```rust
/// Delegates to [`Number::compare_via_host`] so that operators (`<`, `<=`, `.max()`, `.sort()`,
/// ...) all use rippled's `Number`  comparison.
impl PartialOrd for Number {
    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
        Some(self.cmp(other))
    }
}

impl Ord for Number {
    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
        // A `Number` only ever holds host-canonicalized bytes, so comparing two valid values
        // cannot fail in practice.
        self.compare_via_host(other).unwrap()
    }
}
```

xrpl-common-stdlib/src/types/number.rs
```diff
@@ -153,6 +154,124 @@ impl Number {
...
+        let mut out = [0u8; NUMBER_SIZE];
+        let rescode = unsafe {
+            host::float_add(
+                self.0.as_ptr(),
+                self.0.len(),
+                other.0.as_ptr(),
+                other.0.len(),
+                out.as_mut_ptr(),
+                NUMBER_SIZE,
+                TO_NEAREST,
```
sappenin: I think we need to elevate this rounding-mode choice to the function itself (e.g., into `float_add` and the other functions).

sappenin (changes requested):

xrpl-common-stdlib/src/types/iou_number.rs
kuznetsss: Reference to `IOUNumber` will be only 4 bytes. I think rule of thumb here is to use `Copy` when type size is less or equal to the reference size.

TimothyBanks: Is there a consistency argument to be made here?  Or would this sort of thing not be confusing to seasoned rust developers in that one number type is reference but another number type is copy?

sappenin: > Is there a consistency argument to be made here? 

I think both Number types are larger than 4 bytes, so neither one should derive `Copy`. Looks like the PR has been updated as such, so I'll resolve this one.

xrpl-common-stdlib/src/types/number.rs
sappenin: Looks like this update was made. Resolving.

sappenin (approved):

xrpl-common-stdlib/src/host/mod.rs
```diff
@@ -34,6 +34,27 @@ pub const FLOAT_ROUNDING_MODES_DOWNWARD: i32 = 2;
 #[allow(unused)]
 pub const FLOAT_ROUNDING_MODES_UPWARD: i32 = 3;
 
+/// Rounding mode for float operations, matching rippled's `Number::RoundingMode`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(i32)]
+pub enum RoundingMode {
```
sappenin, line 32: Not a blocker for this PR, but maybe for tracking via an issue: Should this enum simply hold the numberic value internally? I can't think of a reason to have the enum values living in constants like `host::FLOAT_ROUNDING_MODES_TO_NEAREST` etc.

sappenin (approved):

sappenin (approved):

xrpl-common-stdlib/src/types/number.rs
```diff
@@ -154,6 +153,124 @@ impl Number {
             )
         })
     }
+
+    /// Compares this `Number` to another via the host, backing the [`Ord`]/[`PartialOrd`] impls.
+    fn compare_via_host(&self, other: &Number) -> Result<core::cmp::Ordering> {
```
kuznetsss, line 158: Not a blocker for this PR. Should we have a naming convention for functions and methods using host functions internally? For example, `host_` prefix or `via_host` suffix?

kuznetsss (approved):

xrpl-common-stdlib/src/types/number.rs
ricky122-5: going to add to standup notes, let's discuss

kuznetsss (approved):

PR #275, opened 2026-08-11
fix: Get CI passing

mvadari:

This PR:
* Updates the Docker hash to the latest version of `ripple/se/supported`
* Fixes the host function audit script to avoid capturing comments
* Adds support for the `i32` type in the ledger object generation script
* Runs the ledger object generation script to catch all the latest updates from xrpld

### Context of Change

CI is failing and we shouldn't have it fail

### Type of Change

- [x] Build / CI / tooling change


pwang200 (approved):
Thanks! lgtm.

Issue #132, opened 2026-02-04
Generate host_function derivatives using Rust instead of JS


sappenin:
[This PR](https://github.com/ripple/xrpl-wasm-stdlib/pull/131) introduced more extensive changes to the code-gen of host function derivatives that is quite useful in ensuring we have a single source of truth for host function definitions. 

That said, given this is a Rust project, it's preferable that any codegen be implemented in Rust so that current and future maintainers of this project (1) don't need to have knowledge around multiple languages to maintain the project well and (2) the codegen can more naturally be integrated into the cargo build (ie, remove as many scripts as possible). 

This ticket tracks that work.

As an example, we can consider this code as a replacement of the JS code for generating the current three host function derivatives that could be added to `xrpl-wasm-stdlib/build.rs` (note this code is not well tested, but is supplied here as a starting point):

`xrpl-wasm-stdlib/build.rs`

```rust
use quote::{format_ident, quote};
use std::fs;
use std::path::Path;
use syn::{parse_file, FnArg, Item, ItemTrait, ReturnType, TraitItem, Type};

#[derive(Debug, Clone)]
struct TraitMethod {
    name: String,
    params: Vec<Param>,
    return_type: String,
}

#[derive(Debug, Clone)]
struct Param {
    name: String,
    ty: String,
}

fn main() {
    println!("cargo:rerun-if-changed=src/host/host_bindings_trait.rs");

    let trait_path = "src/host/host_bindings_trait.rs";
    let trait_content = fs::read_to_string(trait_path)
        .expect("Failed to read host_bindings_trait.rs");

    let ast = parse_file(&trait_content).expect("Failed to parse trait file");
    let methods = extract_trait_methods(&ast);

    println!("Found {} methods in HostBindings trait", methods.len());

    // Generate the three derived files
    generate_wasm_bindings(&methods);
    generate_empty_bindings(&methods);
    generate_test_bindings(&methods);

    println!("✅ Successfully updated function signatures in 3 files");
}

fn extract_trait_methods(ast: &syn::File) -> Vec<TraitMethod> {
    let mut methods = Vec::new();

    for item in &ast.items {
        if let Item::Trait(ItemTrait { ident, items, .. }) = item {
            if ident != "HostBindings" {
                continue;
            }

            for trait_item in items {
                if let TraitItem::Fn(method) = trait_item {
                    let name = method.sig.ident.to_string();
                    
                    let mut params = Vec::new();
                    for input in &method.sig.inputs {
[...]

PR #281, opened 2026-08-19
Migrate tools/ JS scripts to Rust

svirdi-ripple:
## Summary

Ports the six Node.js helper scripts in `tools/` to Rust binaries. Each Rust binary is a drop-in replacement for its JS counterpart, and the `tools/` crate is added as a workspace member so all six build via `cargo`.

New binaries:
- `compare-gas-results`
- `copy-rippled-fixture`
- `compare-host-functions`
- `generate-sfields`
- `generate-test-host-bindings`
- `gas-benchmark`

Addresses #132 — `generate-test-host-bindings` produces byte-identical output to `generateTestHostBindings.js`, satisfying the ticket's core requirement.

Unit tests cover the pure helpers in four of the tools; the codegen tools were also diff-verified end-to-end against their JS versions on real rippled sources.

## Test plan
- [ ] `cd tools && cargo test` passes for all binaries
- [ ] `cargo run --bin generate-test-host-bindings` produces no diff on `xrpl-wasm-stdlib/src/host/host_bindings_{wasm,empty,test}.rs`
- [ ] `cargo run --bin generate-sfields <rippled-src>` produces byte-identical output to `node tools/generateSFields.js <rippled-src>`
- [ ] `cargo run --bin compare-host-functions <rippled-src>` reports the same drift as the JS version
- [ ] `cargo run --bin compare-gas-results` produces byte-identical `GAS_BENCHMARK_REPORT.md`


mvadari (changes requested):
Please port each script in a separate PR

svirdi-ripple:
Closing per @mvadari's review request — will open a separate PR for each tool. First one (compareGasResults → Rust) coming shortly, based off latest main.

PR #287, opened 2026-08-27
fix: Type memo data as StandardBlob in examples

### LINKED ISSUE #203: AtomicSwap Example `get_first_memo` returns incorrect type
In the AtomicSwap example [/atomic_swap1/src/lib.rs]([‎examples/smart-escrows/atomic_swap/atomic_swap1/src/lib.rs‎](https://github.com/ripple/xrpl-wasm-stdlib/pull/202/changes#diff-78ca1886b27a6453ec282cdc8c57f1149e4a578e9a604b8391408e69258efc32)) the function `get_first_memo` seems to return the incorrect type (`ContractData`) when it should instead use a memo-specific type.

`In pub fn get_first_memo() -> Result<Option<(ContractData, usize)>> {`

sappenin:
## Summary
- `get_first_memo` in `atomic_swap1` and `nft_owner` now returns `Result<Option<StandardBlob>>` instead of escrow `ContractData` (and drops the leftover length beside the blob).
- Memo bytes are read via `ctx.tx().path()` on `sfield::MemoData`, matching the field's real type and setting `blob.len` from the host write.
- Skill reference snippets for memo reads were updated to the same typed path.

Fixes #203

## Test plan
- [x] `cargo clippy --manifest-path examples/Cargo.toml --target wasm32v1-none -p atomic_swap1 -p nft_owner -- -Dclippy::all`
- [x] `cargo build --manifest-path examples/Cargo.toml --target wasm32v1-none --release -p atomic_swap1 -p nft_owner`
- [ ] Optional: `./scripts/run-tests.sh examples/smart-escrows/atomic_swap/atomic_swap1` and `examples/smart-escrows/nft_owner` against a local rippled node

### DIFF
--- a/examples/smart-escrows/nft_owner/runTest.js
+++ b/examples/smart-escrows/nft_owner/runTest.js
@@ -1,10 +1,67 @@
 const xrpl = require("xrpl")
 
+async function expectFinishRejected(
+  submit,
+  sourceWallet,
+  escrowResult,
+  memos,
+  label,
+) {
+  const tx = {
+    TransactionType: "EscrowFinish",
+    Account: sourceWallet.address,
+    Owner: sourceWallet.address,
+    OfferSequence: parseInt(escrowResult.sequence),
+    Gas: 1000000,
+  }
+  if (memos !== undefined) {
+    tx.Memos = memos
+  }
+  const response = await submit(tx, sourceWallet)
+  if (response.result.meta.TransactionResult !== "tecBYTECODE_REJECTED") {
+    console.error(
+      `\n${label}: expected tecBYTECODE_REJECTED, got:`,
+      response.result.meta.TransactionResult,
+    )
+    process.exit(1)
+  }
+}
+
 async function test(testContext) {
   const { deploy, finish, client, submit, sourceWallet, destWallet } =
     testContext
 
   const escrowResult = await deploy(sourceWallet, destWallet, finish)
+
+  // Reject before an NFT exists: missing / empty / short / oversize MemoData
+  await expectFinishRejected(
+    submit,
+    sourceWallet,
+    escrowResult,
+    undefined,
+    "absent memo",
+  )
+  await expectFinishRejected(
+    submit,
+    sourceWallet,
+    escrowResult,
+    [{ Memo: { MemoType: xrpl.convertStringToHex("nft_id") } }],
+    "memo without MemoData",
+  )
+  await expectFinishRejected(
+    submit,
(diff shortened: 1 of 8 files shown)


examples/smart-escrows/nft_owner/src/lib.rs
```diff
@@ -3,50 +3,36 @@
...
-        result_code => Err(Error::from_code(result_code)),
-    }
+/// Extracts the first memo from the transaction.
+///
+/// `Memos[0].MemoData` is a `StandardBlob`, not escrow contract data. An empty
+/// or absent memo is `Ok(None)` so the caller can reject it. A missing `Memos`
+/// field used to surface as `FieldNotFound` (a negative host code from this
+/// helper); it is now treated the same as empty, so `escrow_finish` returns `0`
+/// rather than that error code.
+fn get_first_memo(tx: &impl TransactionCommonFields) -> Result<Option<StandardBlob>> {
```
TimothyBanks, line 24: Huge nit:  I get that these are examples programs.  Is there any desire to create a shared library crate to put common functionality for the examples in?

examples/smart-escrows/nft_owner/runTest.js
```diff
@@ -1,11 +1,82 @@
 const xrpl = require("xrpl")
 
+async function expectFinishRejected(
```
TimothyBanks, line 3: Following the similar vein to that nit above, these tests also look like 98% the same.  Would there be an appetite for reducing the code duplication?  I get that these are examples.

TimothyBanks (approved):

TimothyBanks:
There is a test failure that might need to be examine and some merge conflicts.

examples/smart-escrows/nft_owner/src/lib.rs
sappenin: So, I don't think we want a shared crate just for examples. The primary reason is that if examples are doing something a lot, then we should simply add any helper functions directly into the actual stdlib itself.

In this case, I think adding a small helper to TransactionCommonFields trait is the way to go. I've updated this PR via da2a552610209796510c720dbaf8e0ad9ce50e76 and 7f58c43. Let me know what you think.

TimothyBanks (approved):
Unfortunately, there are some more merge conflicts.  But looks good to me.

sappenin:
@TimothyBanks Can you take another look? I worked out the merge conflicts.

TimothyBanks (approved):

PR #290, opened 2026-08-27
test: Trace NFT fields in host_functions_test

### LINKED ISSUE #58: Add e2e-test contract that traces all NFT fields
Similar to the `trace_escrow_ledger_object` e2e-test (see [here](https://github.com/ripple/craft/tree/main/projects/e2e-tests/trace_escrow_ledger_object)), we should create an additional e2e-test that traces all NFT fields to prove that the std-lib properly supports those fields.

Note that this should be distinct from the [nft_owner](https://github.com/ripple/craft/blob/main/projects/nft_owner/src/lib.rs) example because that is an example, whereas this ticket describes a new e2e-test.

sappenin:

Extend `e2e-tests/host_functions_test` so it mints a real NFToken and traces every typed `NFToken` getter (`flags`, `transfer_fee`, `issuer`, `taxon`, `token_sequence`, `uri`) instead of probing `nft_uri` with a dummy all-zero ID.

Fixes #58.

### Context of Change

#58 asked for e2e coverage that proves the stdlib NFT field accessors work against a real host. A dedicated `trace_*` crate would have duplicated `host_functions_test`; this PR folds that coverage into the existing probe and has `runTest.js` mint a token with known flags, taxon, transfer fee, and URI, then pass the ID in an EscrowFinish memo (`StandardBlob`).

Value assertions run only on wasm, because native coverage still uses empty host stubs.

### Type of Change

## Test Plan

- [x] `cargo clippy --manifest-path e2e-tests/Cargo.toml -p host_functions_test --target wasm32v1-none -- -Dclippy::all`
- [x] `cargo test --manifest-path e2e-tests/Cargo.toml -p host_functions_test --lib`
- [x] `cargo build --manifest-path e2e-tests/Cargo.toml -p host_functions_test --target wasm32v1-none --release`
- [ ] `./scripts/run-tests.sh e2e-tests/host_functions_test` (needs a rippled node; local `ws://127.0.0.1:6006` was not running when this was written)

### DIFF
--- a/e2e-tests/host_functions_test/src/lib.rs
+++ b/e2e-tests/host_functions_test/src/lib.rs
@@ -33,10 +33,20 @@ use xrpl_common_stdlib::host::trace::{
 use xrpl_common_stdlib::sfield;
 use xrpl_common_stdlib::types::account_id::AccountID;
 use xrpl_common_stdlib::types::amount::Amount;
+use xrpl_common_stdlib::types::blob::StandardBlob;
 use xrpl_common_stdlib::types::currency::Currency;
 use xrpl_common_stdlib::types::iou_number::IOUNumber;
 use xrpl_common_stdlib::types::mpt_id::MptId;
+use xrpl_common_stdlib::types::nft::{NFT_ID_SIZE, NFToken, flags as nftoken_flags};
 use xrpl_escrow_stdlib::current_tx::escrow_finish::EscrowFinish;
+
+/// Values `runTest.js` puts on the minted NFToken so this contract can check the typed getters
+/// against a real host, not just that they return *something*.
+const EXPECTED_NFT_TAXON: u32 = 12345;
+const EXPECTED_NFT_TRANSFER_FEE: u16 = 1000;
+const EXPECTED_NFT_FLAGS: u16 =
+    nftoken_flags::BURNABLE | nftoken_flags::ONLY_XRP | nftoken_flags::TRANSFERABLE;
+const EXPECTED_NFT_URI: &[u8] = b"https://example.com/nft-metadata.json";
 
 #[unsafe(no_mangle)]
 pub extern "C" fn escrow_finish() -> i32 {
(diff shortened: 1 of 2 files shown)


e2e-tests/host_functions_test/src/lib.rs
```diff
@@ -720,6 +709,143 @@ fn test_utility_functions() -> i32 {
...
 
+/// Trace every NFToken field the stdlib exposes, using the NFT minted by `runTest.js`
+/// (ID passed as EscrowFinish `Memos[0].MemoData`).
+///
+/// Error range: -621 to -632.
+fn test_nft_fields() -> i32 {
+    trace("--- NFT fields (typed NFToken API) ---");
+
+    let escrow_finish = EscrowFinish;
+    let memo = match escrow_finish
```
TimothyBanks, line 720: Could this be moved over to the `get_first_memo` function add in this [PR](https://github.com/ripple/xrpl-wasm-stdlib/pull/287)?

e2e-tests/host_functions_test/src/lib.rs
```diff
@@ -720,6 +709,143 @@ fn test_utility_functions() -> i32 {
...
+        .iter()
+        .rposition(|&b| b != 0)
+        .map(|i| i + 1)
+        .unwrap_or(0);
+    trace_hex("NFT URI:", &uri.data[..uri_end]);
+
+    // Native coverage runs against empty/mock host stubs, which do not reproduce the minted
+    // token. Value checks are only meaningful against rippled.
+    if cfg!(target_arch = "wasm32") {
+        if flags.as_u16() != EXPECTED_NFT_FLAGS {
```
TimothyBanks, line 820: I think using a match instead of the various conditionals would be a bit more idiomatic and clearer.

sappenin: I don't think `match` is the right fit here.  `match` branches on the different forms one value can take. Here we have five independent checks on five different values (flags, transfer fee, issuer, taxon, URI), each bailing out with its own error code (there's no single value to switch on). 

Forcing it into match (a tuple, or match true with guards) would actually be harder to read than the current "if this check fails, return this error" chain, which is the normal idiomatic way to write sequential guard clauses in Rust. That's different from the matches just above, which really are switching over an Ok/Err enum.

If the concern is repetition rather than just the `if` itself, I think we could reduce quite a bit in this file. However, I'd prefer to do that in a different PR. Sound OK?

e2e-tests/host_functions_test/src/lib.rs
sappenin: I think so. Let's work out the "utils" over there first, and then I'll apply that here.

sappenin: Actually, I decided to just update this PR to (see 6abdb63b7ecbf3fe423e9ded103061725373ef85). LMK what you think. Something similar is in #827, so I'll work out the differences depending on which one lands first.

TimothyBanks: Ok.

e2e-tests/host_functions_test/src/lib.rs
TimothyBanks: Ah right, looking at this now, I don't know what I was thinking at the time.

TimothyBanks (approved):

PR #301, opened 2026-09-11
refactor: Migrate compareGasResults.js to Rust

svirdi-ripple:
## Summary

First in a series of per-file migrations of `tools/*.js` → Rust (superseding #281, which was closed per @mvadari's review request to do these one script at a time).

- Adds `tools` as a workspace member with `tools/Cargo.toml` (single binary `compare-gas-results`, `serde` + `serde_json` deps).
- Ports `tools/compareGasResults.js` → `tools/compare_gas_results.rs` (snake_case per `docs/NAMING_CONVENTIONS.md`).
- Wires `scripts/benchmark-gas.sh` to invoke the Rust binary.
- Deletes `tools/compareGasResults.js`.
- Includes unit tests for the pure helpers (`format_number`, `format_int`, `add_thousands_separators`, `append_measurement`).

Verified byte-identical output against a saved JS reference on a fixture. `cargo test --workspace`, `cargo clippy --workspace --all-targets` (with `-Dwarnings`), and `cargo fmt --all --check` all pass.

## Test plan
- [ ] `cargo test --bin compare-gas-results` passes
- [ ] `cargo run --release --bin compare-gas-results` produces byte-identical `GAS_BENCHMARK_REPORT.md` to the previous JS version, given the same `.benchmark/*_results.json` inputs
- [ ] `./scripts/benchmark-gas.sh` still functions end-to-end (Phase 2 now invokes the Rust binary)

## Follow-ups (separate PRs)

The remaining `tools/*.js` scripts will each get their own migration PR:
- `copyRippledFixture.js`
- `compareHostFunctions.js`
- `generateSFields.js`
- `generateTestHostBindings.js`
- `gasBenchmark.js`


Cargo.toml
```diff
@@ -4,7 +4,8 @@ members = [
     "xrpl-common-stdlib",
     "xrpl-macros",
     "xrpl-escrow-stdlib",
-    "xrpl-stdlib-test-utils"
+    "xrpl-stdlib-test-utils",
+    "tools",
```
kuznetsss, line 8: Maybe instead we should have a directory for each tool in case later a tool will get larger than one file?

```suggestion
    "tools/compare_gas_results",
```

tools/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition = "2024"
+publish = false
+
+[[bin]]
+name = "compare-gas-results"
+path = "compare_gas_results.rs"
+
+[dependencies]
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
```
kuznetsss, line 13: It would be better to put all the dependencies to the workspace level
```suggestion
serde.workspace = true
serde_json.workspace = true
```

Cargo.toml
svirdi-ripple: applied in latest commit

tools/Cargo.toml
svirdi-ripple: applied in latest commit

